PR #62, opened 2024-03-15
update defs for core models to v1beta1

MUzairS15:
**Notes for Reviewers**

This PR fixes #


MUzairS15:
@leecalcote Ideally we will be maintaining support for at least 1 older version, and when a version is bumped up, there will be a new directory instead of updating in place. But I think updating in place is ok, as server implicitly supports only the latest.

leecalcote (approved):
Good. 

If it's just as easy to go with `v1alpha2`, we might elect for that considering the other, larger forthcoming changes.

PR #67, opened 2024-03-21
ArtifactHub: Add sample structure for ArtifactHub integration

ctcarrier:
**Notes for Reviewers**

This PR adds a sample definition for integrating Meshery Designs with ArtifactHub

### DIFF
--- a/artifacthub/artifacthub-repo.yml
+++ b/artifacthub/artifacthub-repo.yml
@@ -1,0 +1,5 @@
+# Artifact Hub repository metadata file
+repositoryID: 982dcb40-d869-4712-afd5-526ea860bec7
+owners: 
+  - name: Meshery Maintainers
+    email: maintainers@meshery.io
--- a/artifacthub/designs/kubernetes_basic/1.0.0/artifacthub-pkg.yml
+++ b/artifacthub/designs/kubernetes_basic/1.0.0/artifacthub-pkg.yml
@@ -1,0 +1,20 @@
+version: 1.0.0
+name: kubernetes_basic
+displayName: kubernetes_basic
+createdAt: "2024-03-21T14:19:01Z"
+description: Basic Kubernetes design deploying a service and pod
+license: Apache-2
+homeURL: https://docs.meshery.io/concepts/logical/designs
+appVersion: 1.0.0
+keywords:
+    - kubernetes
+    - meshery
+    - service mesh
+    - orchestration
+links:
+    - name: website
+      url: https://meshery.io/
+    - name: support
+      url: https://discuss.layer5.io/
+provider:
+    name: meshery-maintainers
--- a/artifacthub/designs/kubernetes_basic/1.0.0/README.md
+++ b/artifacthub/designs/kubernetes_basic/1.0.0/README.md
@@ -1,0 +1,1 @@
+Basic Kubernetes design deploying a service and pod
--- a/artifacthub/designs/kubernetes_basic/1.0.0/design.yml
+++ b/artifacthub/designs/kubernetes_basic/1.0.0/design.yml
@@ -1,0 +1,257 @@
+name: kubernetes_basic
+services:
+  AnchorNode:
+    name: AnchorNode
+    type: AnchorNode
+    apiVersion: core.meshery.io/v1alpha1
+    namespace: helloah
+    version: 0.7.1
+    model: meshery-core
+    dependsOn:
+    - helloah
+    traits:
+      meshmap:
+        edges: []
+        id: d9fb6e7b-aa9f-4493-866e-54691a1cc00d
+        label: AnchorNode
+        meshmodel-data:
+          category:
+            metadata: null
+            name: Orchestration & Management
+          components: []
+          displayName: Meshery Core
+          hostID: 00000000-0000-0000-0000-000000000000
+          id: 5400b823-94ee-543c-8877-4ca3e299fcc4
+          metadata:
+            isAnnotation: false
+            published: true
+            subCategory: App Definition and Development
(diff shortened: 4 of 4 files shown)


artifacthub/designs/kubernetes_basic/1.0.0/manifest.yml
```diff
@@ -0,0 +1,257 @@
+name: kubernetes_basic
```
ctcarrier, line 1: Currently when I export a design the file name matches the design name. I think it would simplify the integration with AH if the design file had a consistent name, prefix, or suffix.

artifacthub/designs/kubernetes_basic/1.0.0/artifacthub-pkg.yml
```diff
@@ -0,0 +1,23 @@
...
+    - meshery
+    - service mesh
+    - orchestration
+links:
+    - name: website
+      url: https://meshery.io/
+    - name: support
+      url: https://discuss.layer5.io/
+install: |
+  Does Meshery need to be running in the cluster?
```
ctcarrier, line 21: What would it mean to install a design from AH? Would Meshery need to already be setup in the cluster? Or could someone install a design on a vanilla cluster?

artifacthub/designs/kubernetes_basic/1.0.0/artifacthub-pkg.yml
```diff
@@ -0,0 +1,23 @@
+version: 1.0.0
```
ctcarrier, line 1: I created this manually but ideally this file would be auto-generated. Maybe in some export process?

leecalcote: You are so correct. Currently, Meshery Server does not automatically version designs, leaving this bit to users (for now). Automatic versioning is a desired feature.

artifacthub/designs/kubernetes_basic/1.0.0/artifacthub-pkg.yml
leecalcote: Great question, @ctcarrier. Here's a verbal explanation - 

https://github.com/meshery/schemas/assets/7570704/55de3a1b-ba7f-4168-8930-282d2fc13fcb

The first portion of the recording misses me saying, "Both Artifact Hub and Meshery Cloud Catalog function as central repositories for Meshery Designs...".

Generally, yes, the expectation is that an Artifact Hub user will be running their own Meshery Server, download the design from AH, then import into their Meshery Server.

artifacthub/designs/kubernetes_basic/1.0.0/manifest.yml
leecalcote: Minimally, adding a consistent suffix (filename extension) seems like an excellent suggestion - for a few reasons, yes. 👍

artifacthub/artifacthub-repo.yml
```diff
@@ -0,0 +1,15 @@
...
+# mind that the repository won't be processed if it has not changed since the
+# last time it was processed. Depending on the repository kind, this is checked
+# in a different way. For Helm http based repositories, we consider it has
+# changed if the `index.yaml` file changes. For git based repositories, it does
+# when the hash of the last commit in the branch you set up changes. This does
+# NOT apply to ownership claim operations, which are processed immediately.
+#
+repositoryID: The ID of the Artifact Hub repository where the packages will be published to (optional, but it enables verified publisher)
+owners: # (optional, used to claim repository ownership)
+  - name: Lee Calcote
```
leecalcote, line 14: I wonder if we might be able to use a mailing list like "developers@meshery.io" or if we need named, individual maintainers. Do you know if there are any examples of mailing lists being used, @ctcarrier?

artifacthub/designs/kubernetes_basic/1.0.0/manifest.yml
leecalcote: We should certainly open an issue on meshery/meshery in support of a `.yaml` filename extension for designs that are exported a YAML and a `.tar.gz` filename extension for designs that are exported in OCI format. Would you like to do so, @ctcarrier (open the issue - whether you were to implement or not is also a question for you to consider).

ctcarrier: Sure @leecalcote I'll open an issue. Do you have any preference for the suffix pattern? When I exported the initial YAML it seemed to use whatever name I used in the design. So "Kubernetes Basic.yaml" for instance. So I guess we could change nothing else but add a character delimited suffix for instance:
1. Kubernetes Basic-design.yaml
2. Kubernetes Basic.design.yaml

Or possibly normalize the entire filename to make it snake case or some other format and have the suffix match that format:
1. kubernetes-basic-design.yaml

That seems more invasive and may have other implications. We should be able to follow the same pattern for both `yaml` and `tar.gz` files.

artifacthub/artifacthub-repo.yml
ctcarrier: This page has some information on ownership claims which is the purpose of this section:
https://artifacthub.io/docs/topics/repositories/#ownership-claim

Pertinent to your question I think is: `The user requesting the ownership claim must appear in the list of owners in the metadata file, and the email listed must match with the one used to sign in in Artifact Hub. This information will be used during the process to verify that the requesting user actually owns the repository.`

So I think what matters is that the owner email address matches the email used to login to the AH control panel for Meshery. So maybe since this is early in the process an AH account for an email list could be created to make sure this isn't tied to an individual's email.

leecalcote (approved):

MUzairS15 (approved):
Thanks @ctcarrier

PR #106, opened 2024-07-17
[Chore] Move away from "meshmodel" name

leecalcote:
Signed-off-by: Lee Calcote <lee.calcote@layer5.io>

**Notes for Reviewers**

This PR fixes #


schemas/openapi.yml
```diff
@@ -4799,7 +4799,7 @@ paths:
     get:
       tags:
         - meshery_models
-      operationId: GetMeshModelCategories
```
leecalcote, line 4802: Is this change harmful or need separate consideration?

Yashsharma1911: Not now but as we will use this repo for RTK API generation then yes, RTK hook name is determined by the operationId. I suggest that as we update this here, we also update the operationId in the cloud OpenAPI spec and wherever we use that RTK API in the UI. In future, when we use the schemas repository for RTK generation, this sync update will prevent crash in UI.

PR #164, opened 2024-08-30
update schemas for layers and styles

codeSafari10:
**Notes for Reviewers**

This PR fixes #

- add style properties for component menu  (can be used to drive the config and context menu )
- update the layers schema to be an object , with and array of strings we need to track all of the enabled layers is which is hard 
  to keep in sync in a dynamic design . as new relationships or components get add we need to add them to the layers sections
  and there is not categorization of the layers.  

### DIFF
--- a/schemas/constructs/v1beta1/designs.json
+++ b/schemas/constructs/v1beta1/designs.json
@@ -34,16 +34,11 @@
       "description": "Design-level preferences",
       "properties": {
         "layers": {
-          "type": "array",
-          "description": "List of available layers",
-          "items": {
-            "type": "string"
-          }
+          "type": "object",
+          "description": "List of available layers"
         }
       },
-      "required": [
-        "layers"
-      ]
+      "required": ["layers"]
     },
     "relationships": {
       "description": "List of relationships between components",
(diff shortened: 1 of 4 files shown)


schemas/constructs/v1beta1/designs.json
```diff
@@ -1,63 +1,53 @@
...
+      "description": "Revision of the design as expressed by an auto-incremented, SemVer-compliant version number. May be manually set by a user or third-party system, but will always be required to be of version number higher than the previously defined version number.",
+      "minLength": 1,
+      "maxLength": 50
+    },
+    "components": {
+      "description": "A list of one or more component declarations.",
+      "minItems": 0,
+      "type": "array",
+      "items": {
+        "$ref": "./component.json"
```
leecalcote, line 29: @codeSafari10 do we need to have a version listed here?

leecalcote: @MUzairS15, @aabidsofi19, you might confirm.

aabidsofi19: i think we dont need, because this is pointing to the v1beta1 of components

MUzairS15: The design schema is inside the `v1beta1` and hence `./component.json` points to the `v1beta1` of the components.

MUzairS15 (approved):

PR #191, opened 2024-10-14
Update model struct to contain component and relationship count

Yashsharma1911:
**Notes for Reviewers**

This PR fixes #

### DIFF
--- a/models/v1beta1/model/model.go
+++ b/models/v1beta1/model/model.go
@@ -100,6 +100,14 @@ type ModelDefinition struct {
 	// To prevent cyclic error, the type is changed to interface, it doesn't affect registration
 	// Edited on purpose
 	Relationships interface{} `json:"relationships" gorm:"-" yaml:"relationships"`
+
+	// Total number of components in a model
+	// Edited on purpose
+	ComponentsCount int `json:"components_count" gorm:"-" yaml:"components_count"`
+
+	// Total number of relationships in a model
+	// Edited on purpose
+	RelationshipsCount int `json:"relationships_count" gorm:"-" yaml:"relationships_count"`
 }
 
 // ModelDefinitionMetadataCapabilitiesEntityState defines model for ModelDefinition.Metadata.Capabilities.EntityState.


Yashsharma1911:
// @Jougan-0 FYI

models/v1beta1/model/model.go
```diff
@@ -100,6 +100,14 @@ type ModelDefinition struct {
...
 	// Edited on purpose
 	Relationships interface{} `json:"relationships" gorm:"-" yaml:"relationships"`
+
+	// Total number of components in a model
+	// Edited on purpose
+	ComponentsCount int `json:"components_count" gorm:"-" yaml:"components_count"`
+
+	// Total number of relationships in a model
+	// Edited on purpose
+	RelationshipsCount int `json:"relationships_count" gorm:"-" yaml:"relationships_count"`
```
leecalcote, line 110: use camelCasing

leecalcote (changes requested):

models/v1beta1/model/model.go
```diff
@@ -100,6 +100,14 @@ type ModelDefinition struct {
 	// To prevent cyclic error, the type is changed to interface, it doesn't affect registration
```
leecalcote, line 100: I wonder why the model.go file is under a v1beta1 directory. This is unnecessary and will become antiquated and confusing as the schema progresses in version number.

models/v1beta1/model/model.go
```diff
@@ -100,6 +100,14 @@ type ModelDefinition struct {
...
 	// Edited on purpose
 	Relationships interface{} `json:"relationships" gorm:"-" yaml:"relationships"`
+
+	// Total number of components in a model
+	// Edited on purpose
```
leecalcote, line 105: What does "Edited on purpose" mean?

Yashsharma1911: This file isn't generated from the schema. It's similar to the question I asked on Slack. I believe it was originally generated from the schema and then customized to fit our needs. However, running the schema generation script is overwriting all the changes that were already made to this file.

I suggested updating the JSON schema, as the components and relationship fields in the model file can be handled programmatically during model generation, which would allow us to generate the struct using a script otherwise we are not using benefits of auto struct generation. However, Shlok mentioned that the absence of these fields in model.json is intentional.

I wrote "Edited on purpose" to ensure it is manually updated for now

Yashsharma1911:
I'm moving forward here

PR #221, opened 2025-02-01
Fix: Preserve Field Order in OpenAPI Codegen for Consistent Model Generation

souvikinator:
**Notes for Reviewers**

This PR fixes #

- Moved organization definition inside its package.
- Added `x-order` tag to `v1beta1` constructs to maintain the field order in the generated code.
- Rearranged the fields in construct definition to match the schema.
- Updated reference resolver script to add x-order field to the schemas.

Tests:

- Reference Resolver Script:
   - Add the x-order tag while resolving `$ref`. ✅
   - Handle x-order numbering for nested fields ✅
- Field Rearrangement:
   - After rearranging fields, model generation produces the expected format. ✅
   - Exporting the design in Meshery's design format also outputs as expected. ✅

Next steps:

- Creating a new release for schemas
- Incorporating new releases in Meshery and Meshkit repo.


scripts/ref-resolver.js
```diff
@@ -1,112 +1,129 @@
-var parser = require("@apidevtools/json-schema-ref-parser")
+var parser = require("@apidevtools/json-schema-ref-parser");
```
aabidsofi19, line 1: why are writing this logic ourselves , shouldn't the openapi / jsonschema toolchain automatically resolve the refs

aabidsofi19 (changes requested):

scripts/ref-resolver.js
souvikinator: It was already in place, i just added the x-order logic..must be the formatter showing whole change. 

Answering the question, I tried directly running the codegen it fails to resolve certain refs and has something to do with nested level.

souvikinator: here is the error:
```bash
error generating code: error generating type definitions: error generating Go types for component schemas: error converting Schema model_definition to Go type: error generating Go schema for property 'category': error generating Go schema for property 'id': error turning reference (../core.json#/definitions/uuid) into a Go type: unexpected reference depth: 3 for ref: #/definitions/uuid local: false
```

PR #234, opened 2025-03-12
Resolve ref from the script for typescript schema file

amitamrutiya:
**Notes for Reviewers**

This PR fixes #


amitamrutiya:
@aabidsofi19 

I changed the script; it will resolve the ref as well. Also, because of this, some types files have been changed because now the types are generated from the ref-resolved file. Is that okay? If not, then I will change the logic. I will first generate types before resolving any refs.

aabidsofi19:
@amitamrutiya  only run this before the json generation step , use the original directory for types generation ( that tool does ref resolution internally )  . 
this will otherwise change the types ( remove and add in different files causing breaking changes ).

amitamrutiya:
Done @aabidsofi19. Now all the changes in the types file are from the origin path before ref resolve. I removed some types file because they are not auto generated and not maintained.

aabidsofi19 (approved):

PR #243, opened 2025-03-13
feature: meshery 13849: json model from schemas

n2h9:
**Notes for Reviewers**

This PR implements json model generation for model init command https://github.com/meshery/meshery/issues/13849 

- Add node script to generate json model from schema with `json-schema-default` npm package (no bin comes with package). 
- Update `compile-types.sh` to simultaneously generate json models during typescript types generation. 
- put models inside new folder `json_models`
- update schema.go to be able to get access to json_model folder files when import repo as a library

Comments:
1. `compile-json-schema.mjs` script is implemented a ES modules because importing with require (plain js) is not working for `json-schema-default` as it does not have index.cjs under `node_modules/json-schema-default` 
2. Added a dependency on `lodash.merge` because its merge function doing deep merge and result json have more fields. 
3. new npm packages add no new vulnerabilities. 
4. ~do not see id field in result  `model.json` :woman_shrugging:~ This is resolved by additionaly using `json-schema-ref-parser` :white_check_mark: 


aabidsofi19:
looks good , we might end up changing the target location for these generated default json ( as we are restructuring the schemas ) .

n2h9:
> 4. do not see id field in result model.json :woman_shrugging: 

With respect to this comment in pr description. 
This was because the `json-schema-default` does not resolve the ref fields. So looks like all fields with `$ref` (or maybe without `type`) they were skipped by `json-schema-default`. 

 There is a tool which is doing ref resolving: https://github.com/APIDevTools/json-schema-ref-parser 

This commit: https://github.com/meshery/schemas/pull/243/commits/47fce3c316a4dfa48df7b3da5ea9b1e3f4f6a750 adds ref parser to the script. 
And generated jsons contain all the fields.

go.mod
```diff
@@ -1,6 +1,9 @@
 module github.com/meshery/schemas
 
-go 1.23.0
+go 1.23.4
+
+toolchain go1.24.0
```
aabidsofi19, line 5: these might be accidental from your environment

n2h9: right, :woman_facepalming: good catch, sorry about that, I will update ...

n2h9: I have updated go.mod and go.sum to the version from master as my pr does not touch (almost) go files. 
This:
```sh
go 1.23.4

toolchain go1.24.0
```

appears after run of `go mod tidy` which probably a right way to do :slightly_smiling_face: 
anyway, removing it from this pr. 
thank you :bowing_woman:

PR #250, opened 2025-03-15
Add codegen for Model and Component

aabidsofi19:
**Notes for Reviewers**

This PR fixes #


models/v1beta1/model/model.go
```diff
@@ -77,50 +53,50 @@ type ModelDefinition struct {
...
-	// Edited on purpose
-	ComponentsCount int `json:"components_count" gorm:"-" yaml:"components_count"`
-
-	// Total number of relationships in a model
-	// Edited on purpose
-	RelationshipsCount int `json:"relationships_count" gorm:"-" yaml:"relationships_count"`
+	// RelationshipsCount Number of relationships associated with the model.
+	RelationshipsCount int         `gorm:"-" json:"relationships_count" yaml:"relationships_count"`
+	Components         interface{} `gorm:"-" json:"components" yaml:"components"`
+	Relationships      interface{} `gorm:"-" json:"relationships" yaml:"relationships"`
```
amitamrutiya, line 86: I just have one silly question. Shouldn't this point to the component and relationship schema rather then interface?

aabidsofi19: yeah but the consumers  are right now expecting this to interface so I am not modifying the types

PR #255, opened 2025-03-25
Add new properties for model schema and remove old typescript schema object

amitamrutiya:
I removed some typescript old schema object
I created a new schema for subcategories like we have for the category. If we don't want then I will add into the core schema.
update model schema


schemas/constructs/v1beta1/model/model.json
```diff
@@ -122,20 +135,16 @@
...
     "subCategory": {
-      "type": "string",
-      "description": "Sub-category of the model.",
-      "minLength": 1,
-      "maxLength": 100,
-      "x-oapi-codegen-extra-tags": {
-        "yaml": "subCategory",
-        "json": "subCategory"
-      },
+      "$ref": "../subcategory/subcategory.json",
```
aabidsofi19, line 138: add a x-import-go and type so the generated code wont embed it but import the go struct

schemas/constructs/v1beta1/model/model.json
```diff
@@ -209,6 +220,52 @@
             "json": "svgComplete"
           },
           "x-order": 7
+        },
+        "shape": {
```
aabidsofi19, line 224: why are we adding these

schemas/constructs/v1beta1/subcategory/openapi.yml
```diff
@@ -0,0 +1,9 @@
+openapi: 3.0.0
+info:
+  title: Model API
```
aabidsofi19, line 3: wrong title

schemas/constructs/v1beta1/subcategory/openapi.yml
```diff
@@ -0,0 +1,9 @@
...
+  version: 1.0.0
+paths: {}
+components:
+  schemas:
+    CategoryDefinition:
```
aabidsofi19, line 8: wrong name , should be SubCategoryDefinition

aabidsofi19 (changes requested):

schemas/constructs/v1beta1/subcategory/openapi.yml
amitamrutiya: done

schemas/constructs/v1beta1/model/model.json
amitamrutiya: We have the shape property in model metadata that's why I added it. @aabidsofi19 .
image

amitamrutiya: yeah sourceUri doesn't make sense to have. Let me find different place to add that.

schemas/constructs/v1beta1/model/model.json
```diff
@@ -209,6 +220,52 @@
...
+          "x-order": 9,
+          "x-oapi-codegen-extra-tags": {
+            "yaml": "sourceUri",
+            "json": "sourceUri"
+          },
+          "metadata": {
+            "uiType": "url",
+            "urlValidation": true,
+            "validationHint": "Enter either a git:// GitHub URL or an Artifact Hub URL"
+          }
```
amitamrutiya, line 268: @aabidsofi19 I'm not sure where is the best place to store this information.

aabidsofi19: so you need to define the schema for the import endpoint and  inside the payload for the api add these additional fields and also ref to model definition as needed

PR #282, opened 2025-04-16
Convert openapi files into typescript object

amitamrutiya:
**Notes for Reviewers**

This PR fixes #


amitamrutiya:
@aabidsofi19 i commented some export in this file https://github.com/meshery/schemas/pull/282/files#diff-870b5bcb113047adb34855f3d2234a7964aff167e7f39b96bd6be1bba616ae49 and it reduces the bundle size 60%. If we export everything, the bundle size goes up to 13MB. 

I think since we are exporting openapi files we don't need to export json schema as typescript object. We can ref the openapi schema to json schema and export it from openapi.

aabidsofi19:
> @aabidsofi19 i commented some export in this file https://github.com/meshery/schemas/pull/282/files#diff-870b5bcb113047adb34855f3d2234a7964aff167e7f39b96bd6be1bba616ae49 and it reduces the bundle size 60%. If we export everything, the bundle size can go up to 20MB and more. Currently after commenting bundle size is 1.6 MB.
> 
> I think since we are exporting openapi files we don't need to export json schema as typescript object. We can ref the openapi schema to json schema and export it from openapi.

yeah , the openapi would include the subschemas . so yeah we dont need to put the individual json schemas

aabidsofi19 (approved):

amitamrutiya:
> yeah , the openapi would include the subschemas . so yeah we dont need to put the individual json schemas
we can get rid of individual json schema bundling

@aabidsofi19 to achieve this 
- Each schema has to contain related openapi file
- We have to enhance our schema/construct directory structure, like all the schemas have to be in a directory and contain openapi.yml file, currently some are in a directory and some are not?

PR #300, opened 2025-04-24
Clarify displayName schema description and helperText

zihanKuang:
**Notes for Reviewers**

This PR refines the displayName property helperText in the ModelDefinition schema

### DIFF
--- a/schemas/constructs/v1beta1/model/model.json
+++ b/schemas/constructs/v1beta1/model/model.json
@@ -51,7 +51,7 @@
     },
     "displayName": {
       "description": "Human-readable name for the model.",
-      "helperText": "Model display name should be a friendly name for your model.",
+      "helperText": "Model display name may include letters, numbers, and spaces. Special characters are not allowed.",
       "minLength": 1,
       "maxLength": 100,
       "type": "string",
--- a/typescript/constructs/v1beta1/model/ModelOpenApiSchema.ts
+++ b/typescript/constructs/v1beta1/model/ModelOpenApiSchema.ts
@@ -279,7 +279,7 @@ const schema = {
           },
           "displayName": {
             "description": "Human-readable name for the model.",
-            "helperText": "Model display name should be a friendly name for your model.",
+            "helperText": "Model display name may include letters, numbers, and spaces. Special characters are not allowed.",
             "minLength": 1,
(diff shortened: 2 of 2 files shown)


amitamrutiya:
@zihanKuang https://github.com/meshery/schemas/blob/master/schemas/constructs/v1beta1/model/model.json#L40 this is a right place where you should make changes.

After making changes run 
`make generate-ts`
`make generate-golang`

zihanKuang:
Hi @amitamrutiya 👋,

I wanted to confirm with you the issue I ran into and how I resolved it, just to make sure the changes I made are correct.

### What happened

After modifying `model.json` as suggested, I ran the following:

```bash
make generate-ts
make generate-golang
npm run build
```

However, the build failed with this TypeScript error:

```
typescript/index.ts(54,29): error TS2305: Module '"./openapi"' has no exported member 'webhooks'.
```


### What I tried

1. I reset my working directory to the latest commit:
   ```bash
   git reset --hard HEAD
   ```

2. Removed all generated and unrelated files:
   ```bash
   rm -rf dist typescript/home generate-golang
   ```

3. Regenerated everything:
   ```bash
   make generate-ts
   make generate-golang
   ```

But I still saw the same error. I checked `openapi.d.ts` and noticed that `webhooks` wasn’t actually exported from there.


### How I resolved it

I found that `typescript/index.ts` was still trying to import `webhooks`:

```ts
import { components, paths, webhooks } from './openapi';
```

So I removed `webhooks` from both the import and from the `api` namespace:

```ts
export type Webhooks = webhooks;
```

After making those changes, the build succeeded.


Just to clarify: I only did this to make sure the UI and schema worked together locally.

My questions are:

- Is this the right way to handle the error? My change was only to the `helpText` in `model.json`—I just want to make sure this didn’t break anything.
- I only edited `model.json` and ran the expected make commands, but the build failed—although it had worked before. Do you know why that might be?
- For my PR, should I only include the updated `model.json`? I’ve seen past PRs only commit that file, so I wanted to double-check what’s expected.

Thanks again.😊

amitamrutiya:
@zihanKuang the `webhook` error is not because of your changes. It is in the master. I will check why it is failing.

amitamrutiya (approved):

PR #305, opened 2025-04-26
fix: reg exp in core.json

n2h9:
**Notes for Reviewers**

This PR updates the reg exp in core.json for versionString  (the results of make generate-types is also commited).

Golang does not support "lookahead" construction: `?!`. According to [docs](https://pkg.go.dev/regexp/syntax#Perl) (not listed in supported constructions list).

This is required to be able to use with backend validation library. 

Here is[ a link to go playground](https://go.dev/play/p/epiy3uKPDSQ?v=goprev) with this construction which panics with a message:
```sh
panic: regexp: Compile(`([a-z.])*(?!^/)v(alpha|beta|[0-9]+)([.-]*[a-z0-9]+)*$`): error parsing regexp: invalid or unsupported Perl syntax: `(?!`
```

___

Updated version contains construction: `^(([a-z.])+\/?)*`  which allows such versions as:
`models.meshery.io/v1beta1`
`models.meshery.io/smth.else/v1beta1`

Here is a link to go playground which shows usage of this reg exp: [link](https://go.dev/play/p/OqhJR_1km5I?v=goprev), its output is 
```sh
OK: v1
OK: v1alpha1
OK: v2beta3
OK: v1.custom-suffix
OK: models.meshery.io/v1beta1
OK: models.meshery.io/ser.ver/v1beta1
```

___

there is also same construction is used in v1alpah3 relationships.json not changing it for now, let's see if current update works fine. 

### DIFF
--- a/schemas/constructs/core.json
+++ b/schemas/constructs/core.json
@@ -13,7 +13,7 @@
       "minLength": 2,
       "maxLength": 100,
       "description": "API version of the object",
-      "pattern": "([a-z.])*(?!^/)v(alpha|beta|[0-9]+)([.-]*[a-z0-9]+)*$",
+      "pattern": "([a-z.])*(?!^\/)v(alpha|beta|[0-9]+)([.-]*[a-z0-9]+)*$",
       "example": ["v1", "v1alpha1", "v2beta3", "v1.custom-suffix"]
     },
     "semverString": {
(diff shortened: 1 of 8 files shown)


n2h9:
Not clear if we need this :woman_shrugging: 
let me mark it as draft for now to prevent somebody from merging by chance :smile:

PR #306, opened 2025-04-26
feature: validate document with go lib over json schema

n2h9:
**Notes for Reviewers**

This PR adds validation of document over json schema with golang lib.

This is just to show the case, not intended to be merged yet. 


schemas/constructs/v1beta1/model/model.json
```diff
@@ -247,7 +247,6 @@
       "additionalProperties": true
     },
     "model": {
-      "$ref": "./openapi.yml#/components/schemas/Model",
```
n2h9, line 250: this is not compiles because it is yml, deleted for now to be able to compile schema with go library.

schemas/constructs/core.json
```diff
@@ -13,7 +13,6 @@
       "minLength": 2,
       "maxLength": 100,
       "description": "API version of the object",
-      "pattern": "([a-z.])*(?!^/)v(alpha|beta|[0-9]+)([.-]*[a-z0-9]+)*$",
```
n2h9, line 16: (not about to be merged)

this keep failing with error of invalid reg exp, was not able to resolve, removed for now to be able to compile schema.

schemas/constructs/v1beta1/model/model.json
n2h9: This was resolved by converting .yml on the flight to json when is read by validator.

schemas/constructs/core.json
```diff
@@ -13,8 +13,8 @@
...
-      "example": ["v1", "v1alpha1", "v2beta3", "v1.custom-suffix"]
+      "pattern": "^(([a-z.])+\/?)?v(alpha|beta|[0-9]+)([.-]*[a-z0-9]+)*$",
```
n2h9, line 16: updated version, which looks like very close to original, and is valid with respect to golang reg exp parser.

PR #329, opened 2025-06-05
generate openAPI schema for relationships

FaheemOnHub:
Generated openApi Schema for relationships , as we need them in meshery ui for having a relationship builder - a small UI driven from relationship schema using RJSF to provide a form in which users can define new relationships.

@aabidsofi19 , can u please take a look if this is right , as I am first-time generating schemas , so that I can then publish it to npm and @M-DEV-1 can use it inside meshery-ui. 


aabidsofi19:
i think there is something wrong your local build ( there are local file references pushed ) and also you dont need to define new schemas . the relationships schema is already defined in v1alpha3 and v1beta1 ( in another open pr ) .  the ts files are also generated for it . use the v1alpha3/RelationshipSchema.ts . ( only thing that might need to be update is the export that file from index.ts similar to model schema

FaheemOnHub:
> i think there is something wrong your local build ( there are local file references pushed ) and also you dont need to define new schemas . the relationships schema is already defined in v1alpha3 and v1beta1 ( in another open pr ) . the ts files are also generated for it . use the v1alpha3/RelationshipSchema.ts . ( only thing that might need to be update is the export that file from index.ts similar to model schema

yes i saw that localfile , but there was another-same one also added by someuser, so i thought (will be answer in review).

So, should we wait fot those existing pr to get merge or use the predefined RelationshipSchema as import and export them and publish to npm.
Screenshot 2025-06-05 at 12 50 14 PM

FaheemOnHub:
closed as completed by #330

PR #334, opened 2025-06-24
feature: schema driven connections: part 00

n2h9:
**Notes for Reviewers**

This PR  updates connection schema and generated golang model so that later it could be used from meshery server instead of its Connection model. 

Done:
- copied connection schema under connection subfolder (as it is done for others, and it requires also its own openapi.yml so better to have it in subfolder) 
- marked most fields as required, so that in go struct they are not nilable;
-  updated for go names *Id  fields to *ID;
- extracted environment entity from openapi to json schema file; 

This is probably it for this pr, will do further enhancement to make connection and environment usable as meshery connection model in the further pr; 

As couple fields were updated here, to use this version from this pr it will require this changes on meshery side: https://github.com/meshery/meshery/pull/15242 

also will need to bump the version of schema to latest on:
- meshery
- meshkit

Didn't find usage in other repositories (including layer5io org).
there is one usage of Environment generated typescript EnvironmentDefinitionV1Beta1OpenApiSchema in meshery-cloud, but I didn't regenerate this typescript so should be ok (an also environment schema haven't changed). 

TODO:
- :white_check_mark:  check connection where it could be used and remove from v1beta1 root folder (as it is copied to subfolder)
- add Environment schema and subtypes
- - schema :white_check_mark:  subtypes :no_entry: I left them in openapi.yml (for now or permanently);
- :no_entry: add ConnectionPage, ConnectionPayload and others (will do in further pr); 


generate-golang.sh
```diff
@@ -50,6 +50,14 @@ generate_schema_models() {
     # npx @redocly/cli join $merged_output schemas/merged_openapi.yml -o schemas/merged_openapi.yml --without-x-tag-groups
 
     # rm -f "$merged_output"
+
+    if [ $(which gofmt) ]; then
```
n2h9, line 54: this is not particularly related...
if we keep this we will need to regenerated all models in a separate pr.

models/v1beta1/connection/connection_helper.go
```diff
@@ -41,14 +41,14 @@ func (h *Connection) Create(db *database.Handler) (uuid.UUID, error) {
...
 		if err != nil {
 			return uuid.UUID{}, err
 		}
-		return h.Id, nil
+		return h.ID, nil // TODO make field non-nil
 	}
 
 	// else return the id of the existing connection
-	return connection.Id, nil
+	return connection.ID, nil // TODO make field non-nil
```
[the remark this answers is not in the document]
n2h9: I agree, I have made fields not nil and didn't delete the comment, will update.

schemas/constructs/v1beta1/connection/connection.json
```diff
@@ -0,0 +1,137 @@
...
+  "type": "object",
+  "required": [
+    "id",
+    "name",
+    "credential_id",
+    "type",
+    "sub_type",
+    "kind",
+    "metadata",
+    "status"
```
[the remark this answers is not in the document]
n2h9: yes, this is intentional, as we are doing struct same to https://github.com/meshery/meshery/blob/master/server/models/connections/connections.go 
and there it is nilable

aabidsofi19:
@n2h9  this looks good  .

generate-golang.sh
n2h9: I will probably move it to separate pr.

n2h9: This one probably not needed at all, without it generated code is formatted :slightly_smiling_face:

PR #348, opened 2025-07-14
update: default value for schemaVersion for model, component, relathionship

n2h9:
**Notes for Reviewers**

This PR updates default values.
Because this is the format that is expected by model import command.

reference: https://github.com/meshery/meshkit/blob/9ad0b7052cc5651e29ab413b38f85efb0edf1b17/utils/utils.go#L464 

without one of this prefixes, validation returns error

(I will prepare auto generated code in a separate pr).


leecalcote (approved):
Makes sense.

PR #532, opened 2026-01-08
cleanup intermediate merged/resolved schemas

aabidsofi19:
**Notes for Reviewers**

This PR fixes #


generate-golang.sh
```diff
@@ -87,14 +88,15 @@ generate_schema_models "organization" "v1beta1"
 generate_schema_models "connection" "v1beta1"
 generate_schema_models "invitation" "v1beta1"
```
leecalcote: @aabidsofi19 why are we doing this when the schema itself contains its version number?

Or why are we doing this when the directory structure contains the version number?

Either of these two other methods will reduce maintenance.

README.md
leecalcote: @aabidsofi19, don't forget to update the explanation this readme includes about the directory structure. Given that the structure is changing, this portion needs updated.

schemas/openapi.yml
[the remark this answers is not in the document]
aabidsofi19: these are now put inside a _openapi_build folder , which is not committed to git now . i will update these in docs

PR #576, opened 2026-01-15
feat(connection): add status_summary to ConnectionPage

leecalcote:
## Summary

This PR adds a `status_summary` field to the `ConnectionPage` schema to provide aggregate counts of connections grouped by status.

### Changes
- Added `status_summary` property to `connection_page.yaml`
- Property is a map of status string to count (e.g., `{"connected": 5, "disconnected": 2}`)
- Optional field (omitted when not populated)

### Use Case
This enables the `GET /api/integrations/connections` endpoint to return status summary data inline, eliminating the need for a separate status endpoint while providing the same functionality.

### Example Response
```json
{
  "connections": [...],
  "total_count": 10,
  "page": 0,
  "page_size": 10,
  "status_summary": {
    "connected": 5,
    "disconnected": 2,
    "discovered": 3
  }
}
```

### DIFF
--- a/schemas/constructs/v1beta1/connection/connection_page.yaml
+++ b/schemas/constructs/v1beta1/connection/connection_page.yaml
@@ -42,6 +42,7 @@ properties:
     description: Aggregate count of connections grouped by status
     additionalProperties:
       type: integer
+    x-go-type: map[ConnectionStatus]int
     x-oapi-codegen-extra-tags:
       json: "status_summary,omitempty"
       yaml: "status_summary,omitempty"


schemas/constructs/v1beta1/connection/connection_page.yaml
```diff
@@ -37,3 +37,12 @@ properties:
       json: "page_size"
       yaml: "page_size"
     x-order: 4
+  status_summary:
+    type: object
+    description: Aggregate count of connections grouped by status
+    additionalProperties:
+      type: integer
+    x-oapi-codegen-extra-tags:
```
leecalcote:
```suggestion
    x-go-type: map[ConnectionStatus]int
    x-oapi-codegen-extra-tags:
```

PR #610, opened 2026-02-12
Addition of absent meshery server API's 

YASHMAHAKAL:
**Notes for Reviewers**

This PR fixes #
 Addition of absent constructs and refining meshery rest api's 


schemas/constructs/v1beta1/connection/api.yml
```diff
@@ -179,6 +154,64 @@ paths:
...
+        "200":
+          description: Connection status
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ConnectionsStatusPage"
+        "500":
+          description: Server error
+
+  /api/integrations/connections/{connectionKind}:
```
lekaf974: looks this is not valid current server only support get by connection ID

YASHMAHAKAL: Thanks, fixing

ritzorama:
@YASHMAHAKAL merge conflicts. 👀

schemas/constructs/v1beta1/model/api.yml
```diff
@@ -6,7 +6,7 @@ paths:
   /api/meshmodels/register:
     post:
       summary: Register mesh models
-      operationId: RegisterMeshmodels
+      operationId: idRegisterMeshmodels
```
lekaf974, line 9: Why `id` prefix ?

PragalvaXFREZ:
Thanks for the effort here, @YASHMAHAKAL.

Firstly, please resolve all the merge conflicts so that we have a clean starting point to review this PR.

A few things I have noticed so far:

* Artifacts are being pushed into the PR, and those need to be removed.
* This PR does not seem to follow the dual-schema approach taken elsewhere.
* The `x-annotation` here follows the previous behavior, which will likely make the build fail now.
* The newly added endpoints also need to be accounted for in the remote provider.

I am not sure yet whether we can move forward with this PR as-is, but getting it fixed up and reviewable would still be appreciated.

CC: @rishiraj38, @NSTKrishna, this PR could be a useful reference to solidify some of the concepts we discussed earlier, around what works well and what needs to be avoided.

leecalcote:
Invalid

PR #761, opened 2026-04-17
fix(user): tolerate non-UUID selectedOrganizationId in Preference.Scan

alexquincy:
## Summary

- Harden `Preference.Scan` (`models/v1beta1/user/helpers.go`) so a legacy or malformed `selectedOrganizationId` in the JSONB source does not fail the entire scan with `invalid UUID length: 0`.
- Drop the key from the intermediate map when `uuid.Validate` rejects it; `SelectedOrganizationId` stays at its zero value and the next preference write repopulates it normally.
- Adds `helpers_test.go` covering empty string, non-UUID string, and the valid-UUID happy path.

## Why

`SelectedOrganizationId` is typed as a required `uuid.UUID` (non-pointer, non-nullable). `Preference.Scan` calls `MapToStruct`, which JSON-unmarshals the map into the struct — any non-UUID string under that key fails unmarshal with `invalid UUID length: 0`, and the whole preference column is unreadable.

Observed impact in layer5io/meshery-cloud: `UsersDao.GetUser` surfaces the scan error during the Hydra login_challenge leg of OAuth and users are redirected to `/error`. The legacy data was created (or carried over) by the v1.0.1 rename (`selectedOrganizationID` → `selectedOrganizationId`, see meshery-cloud migration `20260401000000_rename_selectedOrganizationID_pref_key`), which didn't filter empty/invalid values.

meshery-cloud has a companion data-cleanup migration (layer5io/meshery-cloud#5044); this patch is the durable fix so the same class of bad input cannot re-break sign-in from any other writer or export.

## Test plan

- [x] `go test ./models/v1beta1/user/ -run TestPreferenceScan` — 3 passing
- [ ] CI green
- [ ] Downstream: bump `github.com/meshery/schemas` in layer5io/meshery-cloud `go.mod`, verify OAuth sign-in succeeds against a DB that still contains the bad data (without running the meshery-cloud migration)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,7 @@ go 1.25.5
 require (
 	github.com/getkin/kin-openapi v0.133.0
 	github.com/gofrs/uuid v4.4.0+incompatible
+	github.com/google/uuid v1.6.0
 	github.com/lib/pq v1.12.1
 	github.com/meshery/meshkit v0.8.70
 	github.com/oapi-codegen/runtime v1.3.1
@@ -33,7 +34,6 @@ require (
 	github.com/go-openapi/swag/jsonname v0.25.4 // indirect
 	github.com/gogo/protobuf v1.3.2 // indirect
 	github.com/google/s2a-go v0.1.9 // indirect
-	github.com/google/uuid v1.6.0 // indirect
 	github.com/googleapis/enterprise-certificate-proxy v0.3.7 // indirect
 	github.com/googleapis/gax-go/v2 v2.15.0 // indirect
 	github.com/jackc/pgpassfile v1.0.0 // indirect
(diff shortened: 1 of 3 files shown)


fitzergerald (approved):

alexquincy:
### Executed test plan

- `go test ./... -count=1` — full schemas test suite passes (`validation`, `models/v1beta1/user`, etc.)
- `go test ./models/v1beta1/user/ -run TestPreferenceScan -count=1` — 3/3 pass (empty string, non-UUID string, valid-UUID happy path preserved)

Downstream verification is staged in [layer5io/meshery-cloud#5045](https://github.com/layer5io/meshery-cloud/pull/5045) (schemas v1.0.6 bump); the full server unit suite is green there.

PR #895, opened 2026-05-13
chore(user): fix profile update schema and field alignment

Junnygram:
# PR Description: Fix User Profile Update Schema and Aligment

## Description
This PR resolves the profile update failure by adding the missing `PUT` mutation to the User API and aligning the field naming conventions with the backend expectations.

### Key Changes
- **API Schema Update:** Added `PUT /api/identity/users/profile` to `v1beta2/user/api.yml` for profile updates.
- **Naming Alignment:** Renamed `firstName` and `lastName` to `first_name` and `last_name` in the User schema to match the snake_case contract required by the backend.
- **Current User Endpoint:** Added a parameterless `GET /api/identity/users/profile` to allow fetching the current user without path construction errors.
- **RTK Hook Generation:** Regenerated `cloud.ts` to include the new `useUpdateUserMutation` hook.
- **Build System Fix:** Downgraded `@apidevtools/json-schema-ref-parser` to a CJS-compatible version to resolve the `ERR_REQUIRE_ESM` error in the bundling script.

## Verification Results
- [x] OpenAPI bundling (`node build/bundle-openapi.js`) completes successfully.
- [x] RTK Query generation (`node build/generate-rtk.js`) produces the `useUpdateUserMutation` hook.
- [x] Schema validation passes.

## Related Issues
- Fixes Layer5 Cloud profile update failure.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "1.2.19",
       "license": "ISC",
       "devDependencies": {
-        "@apidevtools/json-schema-ref-parser": "^15.2.0",
+        "@apidevtools/json-schema-ref-parser": "^9.0.9",
         "@openapi-contrib/openapi-schema-to-json-schema": "^5.1.0",
         "@reduxjs/toolkit": "^2.8.1",
         "@rtk-query/codegen-openapi": "^2.2.0",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -65,7 +65,7 @@
   "author": "",
   "license": "ISC",
   "devDependencies": {
-    "@apidevtools/json-schema-ref-parser": "^15.2.0",
+    "@apidevtools/json-schema-ref-parser": "^9.0.9",
```
Oluwatunmise-olat: @Junnygram Any reason for the downgrade to `apidevtools/json-schema-ref-parser`?

Oluwatunmise-olat (changes requested):

leecalcote (commented):
This needs to be done in coordination with meshery-cloud

Junnygram:
> This needs to be done in coordination with meshery-cloud

all i have access to are the public repo

schemas/constructs/v1beta2/user/api.yml
```diff
@@ -235,28 +260,28 @@ components:
...
-        firstName:
+        first_name:
           type: string
           maxLength: 200
           description: User's first name
           x-oapi-codegen-extra-tags:
             db: "first_name"
-            json: "firstName"
-        lastName:
+            json: "first_name"
```
PragalvaXFREZ: why are we using snake_case ?

PragalvaXFREZ: i think we are running over circles here, the server side needs to be updated rather than the schema as we are moving into canonical camel casing for json tags. Is there anything else I am missing @Junnygram ?

Junnygram: The change to `snake_case` is to align the schema with the **Layer5 Cloud backend (Golang)**, which uses `snake_case` for these fields in its JSON tags and database mappings. Standardizing this in the schema resolves the `400 Bad Request` errors caused by the previous naming mismatch.

PragalvaXFREZ:
@Junnygram , a certain portion of this PR has been covered in this [PR](https://github.com/meshery/schemas/pull/775) as well, please have a look.

package-lock.json
PragalvaXFREZ, line 1: why are the changes here necessary?

Junnygram: The changes in package-lock.json are a result of restoring @apidevtools/json-schema-ref-parser to the latest version (v15.2.0). This update necessitated a refresh of the lock file to ensure dependency integrity.

PragalvaXFREZ: the changes show that it's being restored into an older version ?

Junnygram: PR #775 covers identity listing and preferences, but it's missing the **PUT /api/identity/users/profile** mutation needed for profile updates. .. .. used the old version by mistake, so i restored it to the version that was there

Junnygram: although , the pr can just be closed, to avoid confusion :).. it was because a user raised an issue regarding the profile update

PragalvaXFREZ:
Thanks for the effort, @Junnygram. I’m closing this PR for the following reasons:

1. The issue has already been addressed in PR #775.
2. This PR includes changes unrelated to the original issue.
3. As it stands, the changes are not compatible with meshery-cloud.

After taking a look into the codebase, the root cause of this bug is not casing incompatibility. The actual failure  traced is more likely country / region value shape.

PR #932, opened 2026-06-16
Enchnace connections schema with schema , transitionMap properties and add apis for connection defintiooons

aabidsofi19:
**Notes for Reviewers**

This PR fixes #


schemas/constructs/v1beta3/connection/connection.yaml
```diff
@@ -1,152 +1,233 @@
...
+    enum:
+      - discovered
+      - registered
+      - connected
+      - ignored
+      - maintenance
+      - disconnected
+      - deleted
+      - not found
+  transitionMap:
```
aabidsofi19, line 171: @leecalcote  connection defintioons would now support defining the possible state transitions

schemas/constructs/v1beta3/connection/connection.yaml
```diff
@@ -1,152 +1,233 @@
...
+      db: metadata
+    x-order: 9
+    x-go-type: core.Map
+    x-go-type-import:
+      path: github.com/meshery/schemas/models/core
+    x-go-type-skip-optional-pointer: true
+    type: object
+    description: Schema for the credential Associated with the connection
+
+  connectionSchema:
```
aabidsofi19, line 134: this property is for defining the schema for a particular connection definition

schemas/constructs/v1beta3/connection/connection.yaml
```diff
@@ -1,152 +1,233 @@
...
+      db: metadata
+    x-order: 9
+    x-go-type: core.Map
+    x-go-type-import:
+      path: github.com/meshery/schemas/models/core
+    x-go-type-skip-optional-pointer: true
+    type: object
+    description: Additional connection metadata
+
+  credentialSchema:
```
aabidsofi19, line 123: this property is for defining the schema for credential of a particular connection definition

schemas/constructs/v1beta3/connection/connection.yaml
pontusringblom: Gosh, this is good to extract from Meshery Server.

leecalcote: Yeah, states and transitions supported for the longest time, but locally defined within Meshery Server.

leecalcote:
I'm re-running the schema audit checks - https://github.com/meshery/schemas/actions/runs/27636704780

schemas/constructs/v1beta3/connection/connection.yaml
```diff
@@ -1,152 +1,233 @@
...
+    x-oapi-codegen-extra-tags:
+      db: metadata
```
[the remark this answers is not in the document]
jamieplu: Fixed in bb716120 - credentialSchema now maps to db:"credential_schema" in both connection.yaml and the generated connection.go, resolving the collision with the metadata column.

schemas/constructs/v1beta3/connection/connection.yaml
```diff
@@ -1,152 +1,233 @@
...
+    x-go-type: core.Map
+    x-go-type-import:
+      path: github.com/meshery/schemas/models/core
+    x-go-type-skip-optional-pointer: true
+    type: object
+    description: Schema for the credential Associated with the connection
+
+  connectionSchema:
+    x-oapi-codegen-extra-tags:
+      db: metadata
```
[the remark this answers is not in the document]
jamieplu: Fixed in bb716120 - connectionSchema now maps to db:"connection_schema" in both the source schema and the generated model.

models/v1beta3/connection/connection_definition_helper.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+	c.ID = generatedID
+
+	if c.Model == nil {
+		return uuid.Nil, fmt.Errorf("connection definition model is nil for connection %q (kind %s)", c.Name, c.Kind)
+	}
+
+	mid, err := c.Model.Create(db, hostID)
+	if err != nil {
+		return uuid.UUID{}, err
```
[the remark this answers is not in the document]
jamieplu: Fixed in bb716120 - the Model.Create error path now returns uuid.Nil, consistent with the other returns in this function.

models/v1beta3/connection/connection_definition_helper.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	modelID := core.Uuid(mid)
+	c.ModelID = &modelID
+	err = db.Omit(clause.Associations).Create(&c).Error
```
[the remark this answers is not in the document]
jamieplu: Keeping this consistent with the established pattern: every sibling entity helper passes &receiver to db.Create (component_helper.go:76, model_helper.go:96, relationship_helper.go:61), and that exact path is exercised in production model/component registration, so it works with the pinned GORM version. If we want to move all four helpers to Create(receiver) that's a worthwhile separate repo-wide cleanup rather than a one-file divergence here.

models/v1beta3/connection/connection_definition_helper.go
```diff
@@ -0,0 +1,64 @@
...
+	return entity.ConnectionDefinition
+}
+
+func (c *ConnectionDefinition) GenerateID() (uuid.UUID, error) {
+	return uuid.NewV4()
+}
+
+func (c ConnectionDefinition) GetID() uuid.UUID {
+	return c.ID
+}
```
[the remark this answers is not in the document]
jamieplu: Keeping the value receiver for consistency with every other entity helper (component/model/relationship all use a value-receiver GetID). Switching only this one would diverge from the convention; if pointer receivers are preferred it should be a coordinated change across all entity helpers.

schemas/constructs/v1beta3/connection/api.yml
```diff
@@ -1,689 +1,910 @@
...
+        format: uuid
+
+    connectionDefinitionId:
+      name: connectionDefinitionId
+      in: path
+      required: true
+      description: Connection definition ID
+      schema:
+        type: string
+        format: uuid
```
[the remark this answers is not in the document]
jamieplu: Leaving the inline type/format here to stay consistent with the sibling path parameters in this same file (connectionId, environmentId), which all define the UUID inline. Happy to migrate all of them to the canonical $ref Uuid in a follow-up if preferred.

jamieplu:
## Iteration summary

### CI status clarification
The three red checks (`Blocking schema validation`, `Full schema audit`, `Consumer audit report`) were **canceled** (exit 143 / runner shutdown - concurrency-group cancellation), not genuine validation failures. The two pushes here re-trigger them.

### Schema fix (commit `bb716120`)
`credentialSchema` and `connectionSchema` both carried `db:"metadata"`, colliding with the `metadata` field and each other - three Go fields mapping to one column. Assigned distinct `credential_schema` / `connection_schema` columns in `connection.yaml` and propagated to the generated `connection.go`. Also aligned the `Create` error path with the surrounding `uuid.Nil` convention.

### Cross-repo compile blocker resolved (commit `919843d3`)
`models/v1beta3/connection` did **not compile** - the new helper reports `entity.ConnectionDefinition`, an `EntityType` that did not exist in any MeshKit release. The schema-audit jobs don't compile Go models so they didn't catch it, but `make build` (post-merge artifact generation) and every downstream consumer would have broken.

Resolved via coordinated change:
- **meshery/meshkit#1026** - added `ConnectionDefinition EntityType = "connection"` (following the `ComponentDefinition = "component"` short-noun convention). Merged and **released as v1.0.11**.
- This PR bumps the MeshKit dependency to `v1.0.11`.

**Verified locally:** `go build ./...` passes, `go vet`/`go test` on the connection package pass, `go mod tidy` clean (only the meshkit line changed).

### Review feedback
- ✅ `credentialSchema`/`connectionSchema` db tags - fixed
- ✅ `uuid.UUID{}` -> `uuid.Nil` - fixed
- ↩️ `&c` to `db.Create` and value-receiver `GetID` - kept, since both are the consistent pattern across every sibling entity helper (component/model/relationship) and the `db.Create(&receiver)` path is exercised in production registration. If we want to change either, it should be a coordinated repo-wide cleanup, not a one-file divergence.
- ↩️ inline UUID vs `$ref` for `connectionDefinitionId` - kept inline to match the sibling path params (`connectionId`, `environmentId`) in the same file.

### Follow-on (not in this PR)
Wiring connection definitions through the **full registration pipeline** (the `schemaVersion`->entity mapping in MeshKit's `utils.getEntityType` and the packaging-unit handling in `models/registration/dir.go`) is a larger feature that also needs a packaging-unit field and meshery-side consumption. The MeshKit PR intentionally scoped to the enum constant; the pipeline work should land as a cohesive unit.

jamieplu:
### Update: schema-audit CI is green ✅

Two more commits since the summary above:
- `919843d3` - bump MeshKit to **v1.0.11** (the released `ConnectionDefinition` EntityType), so `models/v1beta3/connection` compiles.
- `933597c1` - **the actual cause of the schema-audit red checks.** The validate-schemas recursive visitors (format-file Rule 42, enum Rule 8, property-constraint Rules 37-41) had no cycle guard. The connection construct's resolved schema graph contains a `$ref` cycle, so those visitors recursed unbounded - the jobs ran for minutes until they were canceled (exit 143), which looked like failures. Added ancestor-path cycle detection that breaks cycles while leaving acyclic traversal unchanged.

All schema-audit jobs now complete and pass:

| Job | Before | After |
|---|---|---|
| Blocking schema validation | hung ~14m → canceled | **pass (48s)** |
| Full schema audit | canceled | **pass (44s)** |
| Consumer audit report | canceled | **pass (1m31s)** |

Verified locally before pushing: `go build ./...`, `go test ./validation/...`, and both validate-schemas variants complete with exit 0 (blocking: no violations; full: pre-existing advisory warnings only).

leecalcote:
@jamieplu thanks for this PR (https://github.com/meshery/meshkit/pull/1026) and MeshKit release (v1.0.11). Please recognize and review the work in-progress from @aabidsofi19 in https://github.com/meshery/meshkit/pull/1025. Upon reflection, what of this work remains necessary / ideal to have?

schemas/constructs/v1beta3/connection/connection.yaml
jamieplu: Thanks - that intent is preserved. The fix in bb716120 gives connectionSchema its own `connection_schema` column; it was previously colliding on `metadata` with both the `metadata` and `credentialSchema` fields, so a connection definition's connection schema now persists independently.

schemas/constructs/v1beta3/connection/connection.yaml
jamieplu: Same intent preserved here - bb716120 moves credentialSchema to its own `credential_schema` column so the per-definition credential schema no longer collides with `metadata`. The two schema fields and `metadata` are now three distinct columns.

schemas/constructs/v1beta3/connection/connection.yaml
jamieplu: Agreed this is a good move - lifting the state machine into the schema (rather than locally defined in Meshery Server) makes it portable and discoverable. For the record the transitionMap shape (`map[string][]ConnectionStateTransition`) validates cleanly and is carried through unchanged by the fixes on this branch; it's part of what the now-passing schema-audit checks cover.

jamieplu:
@leecalcote on [meshery/meshkit#1025](https://github.com/meshery/meshkit/pull/1025) ("Make Connection definitions as first class registry entities") - reviewed. It's the **registration pipeline**, i.e. the third leg of this feature:

- **#932** (this PR) - the connection-definition schema, generated model, and the `entity.Entity` helper.
- **#1026** (merged, released v1.0.11) - just the `ConnectionDefinition` EntityType constant, the minimal unblock so #932 compiles.
- **#1025** - everything that makes a connection definition actually loadable and registrable.

**What's now redundant in #1025 (one line):** its `models/meshmodel/entity/types.go` addition. #1026 already landed `ConnectionDefinition EntityType = "connection"` - **identical name and value** - so after a rebase onto master that line is a duplicate declaration and should just be dropped. No semantic reconciliation needed; the value matches.

**What remains necessary / ideal (keep all of it):**
- `utils.FindEntityType` → `"connections.meshery.io"` mapping.
- `registration/utils.go getEntity` → `case v1beta3.ConnectionSchemaVersion` (correctly consumes the `ConnectionSchemaVersion` const added in #932 - nicely coordinated).
- `PackagingUnit.Connections` field + the `dir.go` / `register.go` wiring that casts and collects them.
- `registry.go` AutoMigrate + Cleanup registration (creates `connection_definition_dbs`).
- `registry/v1beta1/connection_filter.go` - the `entity.Filter` (GetById/Get/Create).

Without #1025 the #932 helper's `Type()`/`Create()` compile but nothing drives them, so #1025 is what turns this from a type into a working registry entity. It should proceed.

**Before #1025 merges:**
1. Rebase onto master and delete the now-duplicate `types.go` const line (bump its MeshKit base to v1.0.11).
2. Its CI is currently red - `lint` and `test` are failing; those need addressing (some may resolve on the rebase, but not all).
3. Confirm the meshery-side consumers of `PackagingUnit` handle the new `Connections` slice, otherwise loaded connection definitions are silently dropped at registration.

I'll leave a short review note on #1025 itself pointing at the rebase + the const overlap.

PragalvaXFREZ:
> ### Update: schema-audit CI is green ✅
> Two more commits since the summary above:
> 
> * `919843d3` - bump MeshKit to **v1.0.11** (the released `ConnectionDefinition` EntityType), so `models/v1beta3/connection` compiles.
> * `933597c1` - **the actual cause of the schema-audit red checks.** The validate-schemas recursive visitors (format-file Rule 42, enum Rule 8, property-constraint Rules 37-41) had no cycle guard. The connection construct's resolved schema graph contains a `$ref` cycle, so those visitors recursed unbounded - the jobs ran for minutes until they were canceled (exit 143), which looked like failures. Added ancestor-path cycle detection that breaks cycles while leaving acyclic traversal unchanged.
> 
> All schema-audit jobs now complete and pass:
> 
> Job	Before	After
> Blocking schema validation	hung ~14m → canceled	**pass (48s)**
> Full schema audit	canceled	**pass (44s)**
> Consumer audit report	canceled	**pass (1m31s)**
> Verified locally before pushing: `go build ./...`, `go test ./validation/...`, and both validate-schemas variants complete with exit 0 (blocking: no violations; full: pre-existing advisory warnings only).

Thank you for the fix and the updates, @jamieplu.

aabidsofi19:
Thanks @jamieplu and @leecalcote for helping with this

PR #935, opened 2026-06-18
[CI] Permissions - Refresh permissons.csv periodically

### LINKED ISSUE #933: [CI] Workflow to refresh permissions.csv
### Current Behavior
Currently, meshery/schemas is the authoritative source of truth for Meshery's API models, Go structs, and TypeScript types. However, permission key management is split across manual updates and redundant workflows:

1. In meshery/schemas: Developers must manually update [build/permissions.csv](https://github.com/meshery/schemas/blob/master/build/permissions.csv) and run generators (make generate-permissions) locally to update:
- [models/permissions/permissions.go (Go constants)](https://github.com/meshery/schemas/blob/master/models/permissions/permissions.go)
- [typescript/permissions.ts (TypeScript constants)](https://github.com/meshery/schemas/blob/master/typescript/permissions.ts)

2. In meshery/meshery:
Workflow: Runs a duplicate, independent GHA workflow [.github/workflows/generate_keys.yml](https://github.com/meshery/meshery/blob/master/.github/workflows/generate_keys.yml) (named Import Keys) to download keys directly from Google Sheets into local files (server/permissions/keys.csv and docs/static/excel/keys.csv).
UI: The React UI manually maintains a duplicate, hardcoded key mapping object in [ui/utils/permission_constants.ts](https://github.com/meshery/meshery/blob/master/ui/utils/permission_constants.ts) instead of consuming the @meshery/schemas package.

### Desired Behavior

1. Single Source of Truth: meshery/schemas automatically pulls new permission keys from Google Sheets via GHA and regenerates the Go/TS constants.
2. No Redundant Importers: Standalone workflows syncing spreadsheet keys in downstream repositories (like meshery/meshery) are retired.
3. Uniform Consumption: 
      - Downstream Go backends import constants directly from github.com/meshery/schemas/models/permissions.
      - Downstream React UIs import constants directly from @meshery/schemas NPM package instead of maintaining duplicate, hardcoded constants files (permission_constants.ts).

---
- 🎨 Wireframes and designs for Meshery UI in [Figma](https://www.figma.com/file/SMP3zxOjZztdOLtgN4dS2W/Meshery-UI)
- 🧪 [Meshery Test Plan Spreadsheet](https://docs.google.com/spreadsheets/d/13Ir4gfaKoAX9r8qYjAFFl_U9ntke4X5ndREY1T7bnVs/edit#gid=0)

PragalvaXFREZ:
**Notes for Reviewers**

Part of: https://github.com/meshery/schemas/issues/933

This PR adds a scheduled GitHub Actions workflow (`.github/workflows/refresh-permissions.yml`) that keeps `build/permissions.csv` in sync with the canonical "Features & Permissions" Google Sheet and regenerates the Go and TypeScript permission constants from it. It replaces the manual hand-editing of `permissions.csv`.

**What it does**

- Runs once daily (cron `41 5 * * *`, 05:41 UTC) and can also be triggered on demand via `workflow_dispatch`, with an optional CSV URL override.
- Fetches the published sheet into `build/permissions.csv`, regenerates `models/permissions/permissions.go` and `typescript/permissions.ts` via `make generate-permissions`, and commits the result directly to `master`.
- Runs only on `meshery/schemas` (forks are skipped) and uses a single concurrency group so two refreshes never race.
- The sheet is treated as the source of truth, so added, renamed, and removed permissions all flow through without any content gate.
- The one safety check is on the fetch itself: if the download is not the expected CSV (for example a redirect or an error page because the sheet was unpublished or the URL rotated), the run stops instead of overwriting `permissions.csv` with garbage.

**Action required before merging**

Once this PR merges, the workflow treats the Google Sheet as the single source of truth and overwrites `build/permissions.csv` on every run. Any manual change made directly to `permissions.csv` (or to the generated Go / TypeScript constants) will be overwritten on the next refresh.

Before merging we must therefore ensure:

- The spreadsheet is fully canonical and reflects the complete, intended set of permissions.
- Any divergence that exists today between `permissions.csv` and the spreadsheet is reconciled in the spreadsheet first. This includes any change that was applied directly to the CSV but never reflected back into the sheet. If this is not done, the first scheduled run will silently revert those changes on `master`.

After merge, the only durable way to change permissions is to edit the spreadsheet. Direct edits to `permissions.csv` are no longer preserved.

**Other notes**

- The generator skips any row whose Key ID is not a valid UUID, so a typo in the sheet does not break generation; that row is simply omitted from the output. Keep the Key ID column clean so permissions are not unintentionally dropped.
- Removing a permission from the sheet removes its generated constant downstream, which is a breaking change for consumers (meshery/meshery, meshery-cloud). Deletions in the sheet should be deliberate.
- The workflow pushes to `master` using the `MESHERY_CI` token.


jamieplu (approved):
Please compare with the approach in meshery/meshery, and meshery-extensions-packages, please.

rishiraj38 (approved):

rishiraj38:
#933

PR #936, opened 2026-06-19
validation: extend Rule 24 security scheme validation

### LINKED ISSUE #721: Validator: validate security scheme types and format consistency
## Context

Rule 24 checks that \`securitySchemes\` exist and are referenced, but does not validate the scheme definition itself.

## Problem

A security scheme declared as \`jwt\` could have \`type: apiKey\` instead of \`type: http, scheme: bearer\`, and the validator would not catch it. Similarly, no validation of \`scopes\` for OAuth2 schemes.

## Proposal

Extend Rule 24 to also validate:
- \`type: http\` schemes have a valid \`scheme\` value (\`bearer\`, \`basic\`, etc.)
- \`type: apiKey\` schemes have \`in\` and \`name\` defined
- \`type: oauth2\` schemes have \`flows\` defined with valid flow objects
- Warn when a scheme named \`jwt\`/\`bearer\` uses a non-\`http\` type

## Impact

Prevents misconfigured auth schemes from reaching generated clients.

## References

Identified in schema validator audit — Section 2b (Partially Covered).

Omkar-Ugal:
Fixes #721 

Rule 24 previously only checked whether security schemes were declared. This PR extends validation for security scheme definitions, including JWT/Bearer, API key, HTTP, and OAuth2 configurations.

Tests added for valid and invalid configurations across all supported scheme types.

Tests
Added test for missing security schemes
Added test for JWT scheme using a non-HTTP type
Added test for Bearer scheme using a non-HTTP type
Added test for a valid JWT bearer scheme

Verification
go test ./validation
go test ./...

### DIFF
--- a/validation/rules_design.go
+++ b/validation/rules_design.go
@@ -310,8 +310,96 @@ func checkRule24(filePath string, doc *openapi3.T, opts AuditOptions) []Violatio
 				RuleNumber: 24,
 			})
 		}
-	}
 
+		if scheme.Type == "apiKey" {
+			if scheme.Name == "" {
+				out = append(out, Violation{
+					File: filePath,
+					Message: fmt.Sprintf(
+						`Security scheme %q of type "apiKey" must define a "name".`,
+						name,
+					),
+					Severity:   classifyDesignIssue(opts),
+					RuleNumber: 24,
+				})
+			}
+
+			if scheme.In == "" {
+				out = append(out, Violation{
+					File: filePath,
+					Message: fmt.Sprintf(
+						`Security scheme %q of type "apiKey" must define an "in" value (header, query, or cookie).`,
+						name,
+					),
+					Severity:   classifyDesignIssue(opts),
+					RuleNumber: 24,
+				})
+			}
+		}
+
+		if scheme.Type == "http" {
+			validSchemes := map[string]bool{
+				"bearer": true,
+				"basic":  true,
+			}
+
+			if scheme.Scheme == "" {
+				out = append(out, Violation{
+					File: filePath,
+					Message: fmt.Sprintf(
+						`Security scheme %q of type "http" must define a scheme.`,
+						name,
+					),
+					Severity:   classifyDesignIssue(opts),
+					RuleNumber: 24,
+				})
+				continue
+			}
+
+			if !validSchemes[strings.ToLower(scheme.Scheme)] {
+				out = append(out, Violation{
+					File: filePath,
+					Message: fmt.Sprintf(
+						`Security scheme %q uses unsupported http scheme %q.`,
+						name,
+						scheme.Scheme,
+					),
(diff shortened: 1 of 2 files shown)


validation/rules_design.go
```diff
@@ -277,13 +277,155 @@ func checkRule24(filePath string, doc *openapi3.T, opts AuditOptions) []Violatio
...
+					RuleNumber: 24,
+				})
+			}
+		}
+
+		if scheme.Type == "http" {
+			validSchemes := map[string]bool{
+				"bearer": true,
+				"basic":  true,
+			}
```
PragalvaXFREZ, line 353: Fine as-is for now, every security scheme in the repo is type: http, scheme: bearer, so this covers all current usage. For reference, kin-openapi (which our loader already runs in loader.go) accepts bearer, basic, negotiate, and digest for type: http. If any of those ever come up, this allowlist will need to grow to match.

Let's add a comment here, explaining this for future reference.

validation/rules_design.go
PragalvaXFREZ: Stale comment it still says the rule just checks that a scheme is declared. It now also validates scheme types, http scheme values, and OAuth2 flows/scopes.

validation/rules_design.go
```diff
@@ -277,13 +277,155 @@ func checkRule24(filePath string, doc *openapi3.T, opts AuditOptions) []Violatio
...
+					Message: fmt.Sprintf(
+						`Security scheme %q of type "apiKey" must define a "name".`,
+						name,
+					),
+					Severity:   classifyDesignIssue(opts),
+					RuleNumber: 24,
+				})
+			}
+
+			if scheme.In == "" {
```
PragalvaXFREZ: we are just checking if scheme.In == " " , so any value like foo can also pass. kin-openapi restricts it to query/cookie/header . We don't have any apiKey schemes so this won't have any impact right now

PragalvaXFREZ (commented):
thanks, i have added some comments, will take a look again after some time

validation/rules_design_test.go
PragalvaXFREZ, line 1: good amount of tests are always appreciated

validation/rules_design.go
PragalvaXFREZ, line 276: Logic's right, but this'll get hard to manage as more checks land. A few small cleanups: work out the severity once at the top instead of per violation (see checkRule28 right below), move validSchemes out of the loop since it's rebuilt each iteration, and pull the repeated Violation{} into one helper. A switch on scheme.Type reads cleaner than the chain of ifs too.

PragalvaXFREZ (commented):
one suggestion, looks good so far.

PragalvaXFREZ:
@Katotodan , given your interest in meshery/schemas, i think reviewing this can be a good starting point for you. Please drop your feedback or any questions that you might have.

PR #947, opened 2026-06-23
feat(user): type per-organization and per-team assigned role names on User

leecalcote:
## What & why

`User.organizations.organizationsWithRoles` and `User.teams.teamsWithRoles` were
declared as opaque object arrays (`items: { type: object }`), so every downstream
consumer received an untyped `object[]` for a user's per-organization and per-team
role assignments. That is exactly the data the Meshery UI needs to display the roles
assigned to the signed-in user within their currently selected organization (e.g. the
error-404 "Current Session" panel), and today it can only be consumed via untyped
client-side parsing.

This PR defines two component schemas — `OrganizationWithRoles` and `TeamWithRoles` —
and references them from the `User` construct so the assigned role names are strongly
typed end to end (Go structs, TypeScript types, and the generated RTK clients).

### Notes
- **Role names are free-form.** `roleNames` is `array<string>` with **no `enum`**: org /
  workspace / team and custom roles are user-generated and created on the fly, so the
  set is intentionally open. The pre-existing **global** `User.roleNames` enum is left
  untouched.
- **In-place augmentation, no version bump.** This tightens existing fields on the
  already-published v1beta2 `user` construct. The wire shape is unchanged (the server
  already emits `{ id, name, …, roleNames }` per membership), so no new API version is
  warranted — only the schema's type declaration becomes precise.
- The field set mirrors what the Meshery Cloud profile endpoint actually emits
  (`OrganizationWithUserRoles` / `TeamWithUserRoles`).

## Generated artifacts
Regenerated with the repo's pinned `oapi-codegen` (v2.5.1) plus the TS/RTK generators.
The diff is scoped to the `user` construct and the `design` construct (which embeds
`User`).

## Test plan
- [x] `node build/bundle-openapi.js` — bundles clean; new schemas inlined into both cloud + meshery outputs
- [x] `oapi-codegen` regen → `go build ./...` + `go test ./models/...` pass
- [x] Generated Go: `OrganizationWithRoles.RoleNames []string` and `TeamWithRoles.RoleNames []string` (free-form, no enum); global `User.RoleNames *[]UserRoleNames` unchanged
- [x] Generated TS/RTK: `organizationsWithRoles[]` / `teamsWithRoles[]` carry `roleNames: string[]`

## Downstream follow-ups (separate PRs)
- **meshery/meshery (UI):** consume the typed `organizations.organizationsWithRoles` from the existing `getLoggedInUser` query in the Current-Session panel; drop the locally-defined `useGetUserOrgRolesQuery`.
- **layer5io/meshery-cloud:** after this is released, bump the schemas dep and consume the generated `OrganizationWithRoles` / `TeamWithRoles` types (it already emits this shape).

## Follow-up: global `roleNames` de-enumerated (commit d975d71)

Review surfaced that the **pre-existing** top-level `User.roleNames` carried a hardcoded 8-value enum, while role names are actually sourced from `array_agg(DISTINCT role_name)` over the `roles` table (a free `varchar(50)`, dynamically populated). The generated `UserRoleNames` closed union was consumed by nobody — meshery-cloud overrides `roleNames` as `pq.StringArray`/`[]string`, meshery never references it — yet it mislabelled the contract and would reject any custom/future role.

Dropped the enum; the global field is now free-form `pq.StringArray`/`string[]`, consistent with the new per-org/per-team fields and the data model. Backward-compatible widening (seeded literals remain valid strings); no version bump.

## `getAllRoles` documented as the canonical role-listing source

Since `roleNames` is now free-form, tooling must enumerate an org's roles dynamically rather than hard-code a set. The `role` construct's `getAllRoles` (`GET /api/identity/orgs/{orgId}/roles`) description is expanded to state it returns the seeded **plus** custom roles and is the canonical, drift-free source for role pickers / dropdown defaults / docs. No new static enum is introduced (that would just recreate the drift). Downstream doc: `meshery-cloud/docs/roles-reference.md`.

### DIFF
--- a/schemas/constructs/v1beta2/user/api.yml
+++ b/schemas/constructs/v1beta2/user/api.yml
@@ -499,12 +499,14 @@ components:
             json: "deletedAt,omitempty"
         roleNames:
           type: array
+          x-go-type: pq.StringArray
+          x-go-type-import:
+            path: github.com/lib/pq
           description: >-
             Names of the roles assigned to the user within this organization.
             Free-form, user-generated role names; not a fixed enumeration.
           items:
             type: string
-          x-go-type-skip-optional-pointer: true
           x-order: 10
           x-oapi-codegen-extra-tags:
             db: "role_names"
@@ -588,12 +590,14 @@ components:
             json: "deletedAt,omitempty"
         roleNames:
           type: array
+          x-go-type: pq.StringArray
+          x-go-type-import:
+            path: github.com/lib/pq
           description: >-
             Names of the roles assigned to the user within this team.
             Free-form, user-generated role names; not a fixed enumeration.
           items:
             type: string
-          x-go-type-skip-optional-pointer: true
           x-order: 9
           x-oapi-codegen-extra-tags:
             db: "role_names"
--- a/schemas/constructs/v1beta2/user/api.yml
+++ b/schemas/constructs/v1beta2/user/api.yml
@@ -382,7 +382,7 @@ components:
               type: array
               description: Team memberships for the user with their assigned roles.
               items:
-                type: object
+                $ref: "#/components/schemas/TeamWithRoles"
               x-oapi-codegen-extra-tags:
                 db: "teams_with_roles"
                 json: "teamsWithRoles"
@@ -404,7 +404,7 @@ components:
               type: array
               description: Organization memberships for the user with their assigned roles.
               items:
-                type: object
+                $ref: "#/components/schemas/OrganizationWithRoles"
               x-oapi-codegen-extra-tags:
                 db: "organizations_with_roles"
                 json: "organizationsWithRoles"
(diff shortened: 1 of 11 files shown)


schemas/constructs/v1beta2/user/api.yml
```diff
@@ -355,19 +355,19 @@ components:
...
-              - user
-          description: List of global roles assigned to the user
-          example: ["admin", "meshmap"]
+          description: >-
+            Names of the global roles assigned to the user. Free-form,
+            user-generated values sourced from the roles table (role_name is a
+            varchar, not a fixed enumeration); the seeded system roles such as
+            "admin", "organization admin" and "user" are a subset, not the whole
+            set.
+          example: ["admin", "organization admin"]
```
leecalcote:
```suggestion
          example: ["organization admin", "user"]
```

PR #975, opened 2026-06-29
fix: add missing schema refs for object arrays across beta versions

### LINKED ISSUE #973: Missing Schema $refs for paginated object arrays in API specs
### Current Behavior

Several endpoints in the REST API documentation (defined in `meshery/schemas`) that return arrays of objects are missing their `$ref` pointers. Instead of referencing the concrete schema model, the `items` property is left as a generic `type: object`. This causes the generated OpenAPI documentation and client SDKs to lack visibility into the actual returned data structures.

Specifically, the following high-priority locations are missing `$ref`s:

1. `EnvironmentConnectionsPage.properties.connections`
2. `MeshModelModelsPage.properties.models` 
3. `User.properties.teams.properties.teams_with_roles` 
4. `User.properties.organizations.properties.organizations_with_roles`
5. `ResourceAccessActorsResponse.properties.users`
6. `ChildNode.properties.children` in Academy 

### Expected Behavior
The `items` array for these properties should explicitly use a `$ref` to point to the canonical OpenAPI component so the documentation accurately reflects the return types. 

### Environment

- **Host OS:** Mac Linux Windows
- **Platform:** Docker or Kubernetes
- **Meshery Server Version:** stable-v
- **Meshery Client Version:** stable-v

---
- 🎨 Wireframes and [designs for Meshery UI](https://www.figma.com/file/SMP3zxOjZztdOLtgN4dS2W/Meshery-UI) in Figma [(open invite)](https://www.figma.com/team_invite/redeem/GvB8SudhEOoq3JOvoLaoMs)

chaitanyamedidar:
**Notes for Reviewers**

This PR fixes #973 

This PR replaces generic `type: object` and `additionalProperties: true` definitions with explicit `$ref` pointers across core domain arrays in the schemas. This bridges a significant documentation gap and ensures strict, strongly-typed API contracts for clients and generators. 

**Key Changes:**

- **Environment** (`v1beta3`): `EnvironmentConnectionsPage.connections` now references the `Connection` schema.
- **Design** (`v1beta1/v1beta2/v1beta3`): `ResourceAccessActorsResponse.users` now references the `User` schema.
- **Model** (`v1beta1/v1beta2`): `MeshModelModelsPage.models` now references the `Model` schema.
- **Academy** (`v1beta1/v1beta2/v1beta3`): `ChildNode.children` now recursively references the `ChildNode` schema.
- **User** (`v1beta1`): `teams_with_roles` and `organizations_with_roles` now reference the `Team` and `Organization` schemas.
- **Build Script**: Fixed a circular JSON resolution crash in `build/bundle-openapi.js` by passing `{ dereference: { circular: 'ignore' } }` to `$RefParser`, preventing infinite loops on recursive schemas like `ChildNode`.

**Additional Note:**
The `$ref` in `EnvironmentConnectionsPage` correctly documents the intended schema contract for `GET /api/environments/{id}/connections`. However, a review of the local provider implementation in `meshery/meshery` (`environment_persister.go`) suggests the SQL query currently only `SELECT`s a subset of Connection fields rather than the full struct. A follow-up issue should be opened in `meshery/meshery` to align the local provider's SQL query with the full `Connection` schema.


schemas/constructs/v1beta1/model/api.yml
```diff
@@ -282,6 +282,5 @@ components:
         models:
           type: array
           items:
-            type: object
-            additionalProperties: true
+            $ref: "#/components/schemas/Model"
```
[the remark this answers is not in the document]
chaitanyamedidar: Accepted. This ensures type consistency with the rest of the schemas and enforces strict Go struct generation.

schemas/constructs/v1beta2/model/api.yml
```diff
@@ -294,8 +294,7 @@ components:
         models:
           type: array
           items:
-            type: object
-            additionalProperties: true
+            $ref: "#/components/schemas/Model"
```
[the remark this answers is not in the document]
chaitanyamedidar: Accepted. This ensures type consistency with the rest of the schemas and enforces strict Go struct generation.

schemas/constructs/v1beta1/user/api.yml
```diff
@@ -370,7 +370,7 @@ components:
               type: array
               description: Team memberships for the user with their assigned roles.
               items:
-                type: object
+                $ref: "../team/api.yml#/components/schemas/Team"
```
hortison: `teams_with_roles` is documented as "Team memberships for the user with their assigned roles," but this now `$ref`s the bare `Team` entity (`team.yaml`), which has no role/role-name field at all (`id, name, description, owner, metadata, created_at, updated_at, deleted_at`). There's nowhere in the new strict shape to carry per-membership role data, so this loses exactly the information the field name promises.

`v1beta2/user/api.yml` already solves this correctly for the identical field pair — it defines a dedicated `TeamWithRoles` schema (line 515) with an `id`, `name`, and `roleNames: string[]` array, and references that instead of the bare `Team`. v1beta1 has no equivalent `TeamWithRoles` schema; one should be added (snake_case `role_names` to match v1beta1's existing wire casing) and referenced here instead of `Team`.

Same issue applies to `organizations_with_roles` below (line 395) against the bare `Organization` schema — v1beta2 solves it with `OrganizationWithRoles`.

schemas/constructs/v1beta1/user/api.yml
```diff
@@ -392,7 +392,7 @@ components:
               type: array
               description: Organization memberships for the user with their assigned roles.
               items:
-                type: object
+                $ref: "../organization/api.yml#/components/schemas/Organization"
```
hortison: Same issue as `teams_with_roles` above: `organizations_with_roles` is described as carrying "assigned roles," but `Organization` (v1beta1/organization/api.yml:488) has no role field either. Should reference a new `OrganizationWithRoles` schema (mirroring v1beta2/user/api.yml's `OrganizationWithRoles` at line 420) instead of the bare `Organization` entity.

schemas/constructs/v1beta1/environment/api.yml
```diff
@@ -146,9 +146,8 @@ components:
         connections:
           type: array
           items:
-            type: object
-            additionalProperties: true
-          description: The connections of the environmentconnectionspage.
+            $ref: "../connection/api.yml#/components/schemas/Connection"
```
hortison: `Connection` (v1beta1/connection/api.yml:353) has `additionalProperties: false` and a non-trivial `required` list (`id, schemaVersion, name, type, subType, kind, status`). This PR's own description already flags that `environment_persister.go`'s SQL query only selects a subset of Connection columns — meaning the real server response for this endpoint likely won't satisfy the new strict `required` list (e.g. `schemaVersion` isn't a DB column). If anything downstream validates responses against this schema (or a strict client deserializer), this endpoint could start failing until the follow-up issue against `meshery/meshery` mentioned in the PR description actually lands. Worth either holding this specific change until the persister is aligned, or filing that follow-up before merge rather than after. Same applies to v1beta3/environment/api.yml:185.

build/bundle-openapi.js
```diff
@@ -178,7 +178,7 @@ function mergeOpenapiSpec(baseSpec, specToMerge) {
 }
 
 async function dereferenceOpenapiSpec(inputPath) {
-  return $RefParser.dereference(inputPath);
+  return $RefParser.dereference(inputPath, { dereference: { circular: "ignore" } });
```
hortison: This is a correct and necessary fix — I reproduced the crash on the old code (`Converting circular structure to JSON`, triggered by the new `ChildNode` self-`$ref`) and confirmed `circular: "ignore"` resolves it, with the full bundle/Go/TS pipeline succeeding end-to-end.

One scoping note: this option is passed unconditionally on the single `dereferenceOpenapiSpec` call site used by `bundleSchema()` for every one of the 36+ discovered packages, not just `academy` (the only package in this PR with a genuine circular ref). `build/lib/config.js` already has a precedented per-package exclusion mechanism (`excludePackages`, with an existing comment about `v1beta1/relationship`'s circular-$ref problem) that this could have used instead. As written, if some other package later gains an accidental circular `$ref` (e.g. a copy-paste mistake), the bundler will now silently leave it un-dereferenced instead of failing the build loudly — trading a build-time signal for a possible silent downstream issue. Not blocking, but worth scoping to `academy` specifically if convenient.

schemas/constructs/v1beta2/design/api.yml
```diff
@@ -1053,8 +1053,7 @@ components:
         users:
           type: array
           items:
-            type: object
-            additionalProperties: true
+            $ref: "../user/api.yml#/components/schemas/User"
           description: The users of the resourceaccessactorsresponse.
```
hortison: Minor: this PR cleans up the equivalent stale, auto-generated-looking description on `environment/api.yml` ("The connections of the environmentconnectionspage." → "The connections associated with the environment.") but leaves this one as-is. Same fix would apply here and at v1beta3/design/api.yml:1257.

```suggestion
          description: The users with access to this resource.
```

schemas/constructs/v1beta3/design/api.yml
```diff
@@ -1253,8 +1253,7 @@ components:
         users:
           type: array
           items:
-            type: object
-            additionalProperties: true
+            $ref: "../../v1beta2/user/api.yml#/components/schemas/User"
           description: The users of the resourceaccessactorsresponse.
```
hortison: Same stale description as v1beta2/design/api.yml:1057.

```suggestion
          description: The users with access to this resource.
```

schemas/constructs/v1beta1/academy/api.yml
```diff
@@ -1374,7 +1374,7 @@ components:
           type: array
           description: List of child nodes (sub-courses or modules)
           items:
-            type: object
+            $ref: '#/components/schemas/ChildNode'
```
hortison: Nit: single-quoted `$ref` here, but this file otherwise uses double quotes for every other `$ref` (137 double-quoted vs. this 1 single-quoted). `v1beta2`/`v1beta3` academy's `api.yml` are fine as-is since single quotes are already their dominant convention there.

```suggestion
            $ref: "#/components/schemas/ChildNode"
```

hortison (commented):
Reviewed and verified against the actual build pipeline (`bundle-openapi.js`, `go run ./cmd/validate-schemas`, `generate-golang.js`, `go test ./...`) — the core mechanism (object → `$ref`, plus the `circular: "ignore"` fix for the new recursive `ChildNode`) is sound and empirically confirmed working. Inline comments cover the issues found in the diff itself. Two more that don't map to a diff line:

**1. `MeshModelModelsPage.models` (v1beta1 & v1beta2 `model/api.yml`) — claimed fixed, but not actually changed.** The PR description says "Model (v1beta1/v1beta2): `MeshModelModelsPage.models` now references the `Model` schema," and issue #973 explicitly lists `MeshModelModelsPage.properties.models` as item 2 of 6. Neither `schemas/constructs/v1beta1/model/api.yml` nor `v1beta2/model/api.yml` appears in this diff — `models.items` is still `type: object` / `additionalProperties: true` in both. The fix is the same one-line pattern used elsewhere in this PR:
```yaml
        models:
          type: array
          items:
            $ref: "#/components/schemas/Model"
          description: The models matching the list-endpoint query.
```
(`Model` is already defined locally in both files.) Worth adding before merge so this doesn't ship as a partial fix for #973 while the description claims it's complete.

**2. No general lint rule for this whole bug class.** `cmd/validate-schemas` has no rule that catches "`items: {type: object, additionalProperties: true}` used as a missing-`$ref` placeholder" in general — the closest (Rule 30 in `validation/rules_design.go`) only inspects top-level response schemas, not nested `components/schemas` refs like these, and only fires at ≥3 declared properties, so it wouldn't have caught these anyway. `grep -rln "additionalProperties: true" schemas/constructs/ | wc -l` still returns 43 files after this PR. Spot-checking found at least one more likely-genuine instance of the same bug: `TeamMember` in `schemas/constructs/v1beta1/team/api.yml` (`type: object`, `additionalProperties: true`, zero declared properties, no free-form-intent description). Not a blocker for this PR, but a rule generalizing this check would stop the remaining instances from silently persisting — worth a follow-up issue.

Nice catch on the circular-ref crash and the six targeted `$ref` fixes overall — the schema documentation gap this closes is real and worth landing once the role-data and Model gaps above are addressed.

schemas/constructs/v1beta1/user/api.yml
chaitanyamedidar: Addressed. v1beta1 now uses a dedicated TeamWithRoles schema for teams_with_roles so the role_names payload is preserved instead of referencing the bare Team entity.

schemas/constructs/v1beta1/user/api.yml
chaitanyamedidar: Addressed. v1beta1 now uses OrganizationWithRoles for organizations_with_roles so assigned role_names remain part of the response shape.

schemas/constructs/v1beta1/environment/api.yml
chaitanyamedidar: Acknowledged. I’m leaving this as the stricter Connection ref from the issue scope, and the downstream persister alignment should be tracked before merge if maintainers want that as a prerequisite.

build/bundle-openapi.js
chaitanyamedidar: Addressed. Scoped circular dereference handling to academy packages instead of applying circular: ignore to every bundled package.

schemas/constructs/v1beta2/design/api.yml
chaitanyamedidar: Addressed. Updated the stale ResourceAccessActorsResponse users description in v1beta2 design.

schemas/constructs/v1beta3/design/api.yml
chaitanyamedidar: Addressed. Updated the same ResourceAccessActorsResponse users description in v1beta3 design.

schemas/constructs/v1beta1/academy/api.yml
chaitanyamedidar: Addressed. Switched the v1beta1 academy ChildNode ref to double quotes to match the file convention.

PR #1040, opened 2026-07-10
Add missing 403 responses to admin-gated user operations

### LINKED ISSUE #1022: User api.yml operations omit 403 from documented responses on admin-gated endpoints
### Current Behavior

Operations in `schemas/constructs/v1beta2/user/api.yml` document 401/404/500 responses but omit 403, even where the serving handler returns it (observed on `getUserEmailAddresses`, whose meshery-cloud handler 403s non-self/non-admin callers; the same sibling-response pattern applies to other admin-gated user operations in the file - the convention was inherited from `getUserProfileById`).

### Expected Behavior

Documented responses match what handlers emit (schema/handler parity). Admin- or self-gated operations should document 403.

### Proposed fix

Sweep the user (and likely org/team) api.yml operation responses for missing 403s on gated endpoints; add a shared 403 response component; regenerate. Low risk: response-set additions are additive.

### Environment
- v1.3.24 @ 2026-07-03 (found during meshery-cloud account-consolidation review)

vanshika921vd:
## Notes for Reviewers

This PR updates the user API schema to document missing `403 Forbidden` responses for admin- and self-gated operations.

### Changes
- Added shared `403` response component.
- Added missing `403` responses to user operations that already documented `401 Unauthorized`.

Fixes #1022

Thank you for contributing to Meshery! 

## Summary by CodeRabbit

* **Documentation**
  * Added documentation for HTTP `403 Forbidden` responses across core, organization, team, and user API endpoints.
  * Standardized the response format with a plain-text error message.
  * Updated endpoint specifications to clearly indicate when access may be denied.


schemas/constructs/v1beta2/user/api.yml
```diff
@@ -77,6 +79,8 @@ paths:
           $ref: "#/components/responses/400"
         "401":
           $ref: "#/components/responses/401"
+        "403":
+          $ref: "#/components/responses/403"
```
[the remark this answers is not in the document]
vanshika921vd: I followed the guidance from issue #1022, which states that endpoints already documenting a 401 Unauthorized response should also document 403 Forbidden. I'll wait for maintainer guidance on whether /api/users should remain public or if the documented responses/security should be adjusted.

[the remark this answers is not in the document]
chaitanyamedidar: @vanshika921vd if an endpoint explicitly declares `security: []` (meaning it's public), it shouldn't even have a `401` response.

chaitanyamedidar (changes requested):
Missing component definition for `403` status code. 
Also, this PR only updates `v1beta2/user/api.yml`. We need to look into the `org` and `team` files as well.

vanshika921vd:
> Missing component definition for `403` status code. Also, this PR only updates `v1beta2/user/api.yml`. We need to look into the `org` and `team` files as well.

@chaitanyamedidar Thanks! I've addressed the review, updated the PR

schemas/constructs/v1beta2/user/api.yml
```diff
@@ -383,6 +397,8 @@ paths:
           $ref: "#/components/responses/400"
         "401":
           $ref: "#/components/responses/401"
+        "403":
```
chaitanyamedidar: Extra whitespaces after `"403":`.

chaitanyamedidar (changes requested):
Public `/api/users` looks good (`security: []`, no `401`/`403`). Org is partially covered. A few blockers remain:

1. **Broken `403` refs** - `user/api.yml` points at `../core/api.yml#/components/responses/403`, and several org ops use `#/components/responses/403`, but neither core nor org defines a `403` component. Add it next to `401` in core, then `$ref` it from user/org/team.

2. **Team not updated** - `team/api.yml` is untouched. The issue asked for user, org and team, add `403` on gated team ops.

3. **Org consistency** - mix of `$ref` and inline `description: Forbidden`. Prefer the shared `$ref` once it exists. Confirm whether `getOrgs` / `getOrgByDomain` need `403`.

schemas/constructs/v1beta2/organization/api.yml
```diff
@@ -67,6 +67,8 @@ paths:
           $ref: "#/components/responses/400"
         "401":
           $ref: "#/components/responses/401"
+        "403":
+          $ref: "#/components/responses/403"
```
jijillery, line 71: **Broken `$ref` - no `403` response is defined in this file.**

The nine new `403` entries here resolve to `#/components/responses/403`, but this file's `components.responses` only defines `400`, `401`, `404`, and `500` - there is no `403`. Every one of these refs dangles, so the construct fails to bundle:

```
Missing $ref pointer "#/components/responses/403". Token "403" does not exist.
```

Add the missing component under `components.responses`, mirroring the siblings that source from the shared v1alpha1 core:

```yaml
    401:
      $ref: "../../v1alpha1/core/api.yml#/components/responses/401"
    403:
      $ref: "../../v1alpha1/core/api.yml#/components/responses/403"
    404:
      $ref: "../../v1alpha1/core/api.yml#/components/responses/404"
```

schemas/constructs/v1beta2/team/api.yml
```diff
@@ -327,6 +327,8 @@ paths:
           $ref: "#/components/responses/400"
         "401":
           $ref: "#/components/responses/401"
+        "403":
+          $ref: "#/components/responses/403"
```
jijillery, line 333: **Broken `$ref` - no `403` response is defined in this file.**

The eight new `403` entries here resolve to `#/components/responses/403`, but this file's `components.responses` only defines `200`, `400`, `401`, `404`, and `500` - there is no `403`. Every one of these refs dangles, so the construct fails to bundle:

```
Missing $ref pointer "#/components/responses/403". Token "403" does not exist.
```

Add the missing component under `components.responses`, mirroring the siblings that source from the shared v1alpha1 core:

```yaml
    401:
      $ref: "../../v1alpha1/core/api.yml#/components/responses/401"
    403:
      $ref: "../../v1alpha1/core/api.yml#/components/responses/403"
    404:
      $ref: "../../v1alpha1/core/api.yml#/components/responses/404"
```

schemas/constructs/v1beta2/user/api.yml
```diff
@@ -370,6 +388,8 @@ components:
       $ref: "../core/api.yml#/components/responses/400"
     401:
       $ref: "../core/api.yml#/components/responses/401"
+    403:
+      $ref: "../core/api.yml#/components/responses/403"
```
jijillery, line 392: **Broken `$ref` - this points to a core file that has no `403`.**

`../core/api.yml` resolves to `v1beta2/core/api.yml`, whose `responses` block defines `200/201/400/401/404/409/500` but **no `403`**. The shared `403` this PR added lives in `v1alpha1/core/api.yml`, which the user construct does not reference. So every user-operation `403` (they all route through this component) dangles and the construct fails to bundle:

```
Missing $ref pointer "#/components/responses/403". Token "403" does not exist.
```

Fix by adding a `403` response to `v1beta2/core/api.yml` (the file this construct already references for 400/401/404/...), e.g. mirroring its `401`:

```yaml
    "403":
      description: Forbidden
      content:
        text/plain:
          schema:
            type: string
```

Alternatively, retarget just this one ref to `../../v1alpha1/core/api.yml#/components/responses/403`, but that makes the user construct's 403 source a different core file than its other responses.

schemas/constructs/v1beta2/user/api.yml
```diff
@@ -358,6 +374,8 @@ paths:
           $ref: "#/components/responses/400"
         "401":
           $ref: "#/components/responses/401"
+        "403":
```
jijillery: Trailing whitespace after `"403":`.

```suggestion
        "403":
```

schemas/constructs/v1beta2/team/api.yml
```diff
@@ -435,8 +443,10 @@ paths:
                 $ref: "#/components/schemas/TeamMembersPage"
         "400":
           $ref: "#/components/responses/400"
-        "401":
+        "401":
```
jijillery: Trailing whitespace was introduced on this otherwise-unchanged line.

```suggestion
        "401":
```

schemas/constructs/v1beta2/user/api.yml
```diff
@@ -46,15 +46,17 @@ paths:
...
+          $ref: "#/components/responses/403"
         "404":
           $ref: "#/components/responses/404"
         "500":
           $ref: "#/components/responses/500"
-
   # /api/users (getUsers) is owned by the v1beta3 user construct, where its
```
jijillery, line 56: Unrelated whitespace churn: this PR removed the blank line that separated this comment from the operation above it, and added a second blank line before `/api/identity/users/profile/{id}:` below. Neither change relates to the 403 work - please restore the original single-blank-line spacing to keep the diff focused.

schemas/constructs/v1beta2/organization/api.yml
```diff
@@ -119,6 +121,8 @@ paths:
                 $ref: "#/components/schemas/OrganizationPage"
         "401":
           $ref: "#/components/responses/401"
+        "403":
+          $ref: "#/components/responses/403"
```
jijillery, line 125: Scope check: issue #1022 asks for `403` on admin- and self-gated operations, but `403` is being added to nearly every operation across the org/team/user constructs, including read/list endpoints like this `getOrganizations`. Do all of these actually return `403`? For any endpoint that is not permission-gated, documenting a `403` is misleading - worth confirming against the handlers before adding it everywhere.

schemas/constructs/v1beta2/organization/api.yml
jamieplu: Fixed - `organization/api.yml` now defines a local `403` in `components.responses` that bridges to `../../v1alpha1/core/api.yml#/components/responses/403`, matching the 400/401/404/500 siblings. `make bundle-openapi` now emits `_openapi_build/constructs/v1beta2/organization/merged-openapi.json` with the `403` fully resolved and no remaining `$ref` tokens.

schemas/constructs/v1beta2/team/api.yml
jamieplu: Fixed the same way as the organization construct - `team/api.yml` gets a local `403` bridging to `../../v1alpha1/core/api.yml#/components/responses/403`, alongside its existing 200/400/401/404/500 entries. The construct bundles cleanly and all eight operation-level refs resolve.

schemas/constructs/v1beta2/user/api.yml
jamieplu: Took the first option: `403` is now defined in `v1beta2/core/api.yml`, mirroring its `401`. Retargeting the single ref at `v1alpha1/core` would have made this the only bridge block in the repo that sources responses from two different core versions - every other construct resolves its whole `components.responses` block against one core - and `v1beta2/core` already carries its own independent copies of 200/201/400/401/404/409/500, so adding `403` there restores parity rather than introducing a cross-version dependency.

schemas/constructs/v1beta2/user/api.yml
jamieplu: Removed.

schemas/constructs/v1beta2/team/api.yml
jamieplu: Removed - that line is back to its original form and no longer appears in the diff.

schemas/constructs/v1beta2/user/api.yml
jamieplu: Restored the original spacing: a single blank line after `getUsersForOrg` and a single blank line between the comment and `/api/identity/users/profile/{id}`. That hunk is out of the diff now.

schemas/constructs/v1beta2/organization/api.yml
jamieplu: Checked this against the meshery-cloud router and handlers. Two corrections to the premise first: this anchor is `getOrg` (`GET /api/identity/orgs/{orgId}`), not `getOrganizations` - and the actual list operation, `getOrgs` on `GET /api/identity/orgs`, was never given a `403` by this PR, so the collection reads were not blanket-covered.

Of the 26 operations that did gain a `403`, 16 are demonstrably permission-gated at the route layer and legitimately return it:

- **organization (6/9)** - `updateOrg`, `deleteOrg`, `addTeamToOrg`, `removeTeamFromOrg`, `addUserToOrg` sit behind `AuthorizationMiddlewareForOrgAdminAndOrgOwner` / `ForOrgOwner` / `ForOrgAdminAndOrgOwnerOrTeamAdminAndOwner`; `deleteUserFromOrg` behind `PermissionCheckMiddleware(IdentityAccessManagementRemoveUserFromOrganization)`. All of those emit `http.StatusForbidden`.
- **team (8/8)** - `createTeam`/`updateTeam` via `PermissionCheckMiddleware`, `deleteTeam` and the membership ops via the team-admin/owner governance middlewares, and the read ops via `SetOrganizationContext`, which 403s on cross-organization access.
- **user (2/9)** - `getUsersForOrg` via `SetOrganizationContext`; `getUserEmailAddresses` emits `403` directly in the handler.

The remaining 10 have no reachable `403` in the checkout I inspected: `createOrg`, `getOrg`, `getOrgPreferences`, `getUserProfileById`, `getUser`, `getUserProfileOverview`, `notifyMentionUsers`, `createAnonymousUserSession`, `deleteUserAccount`, and `getAccountDeletionEligibility` (the last has no route in cloud yet - the contract here is newer than the implementation).

I have deliberately left those in rather than stripping them. Three of them (`getOrg`, `getOrgPreferences`, `getUserProfileById`) are registered on the unauthenticated root echo instance in cloud while this schema declares `security: [jwt]` and already documented `401` before this PR - so the divergence is between the schema and the handlers, and per the source-of-truth rule that is a cloud-side gap to reconcile, not something to resolve by weakening the contract in a bundling fix. The rest are plain-authenticated operations where a `403` is a permitted future state rather than a documented lie. Worth a separate issue to reconcile `security` and the 401/403 surface for the public-in-cloud routes; happy to file it if that reading looks right.

PR #1045, opened 2026-07-12
fix: add cross-construct cache invalidation for environment connections

### LINKED ISSUE #1044: Environment card shows stale "0 Assigned Connections" — cross-construct RTK cache tag mismatch
## Description

After assigning one or more connections to an Environment and saving, the Environment card's "Assigned Connections" count shows `0`, even though the Connections table correctly shows the environment name against each assigned connection.

## Steps to Reproduce

1. Create a new Environment
2. Assign one or more Kubernetes connections to it
3. Save
4. Observe the Environment card — it shows "0 Assigned Connections"
5. Check the Connections table — each connection correctly lists the environment under its "Environments" column

## Expected Behavior

The Environment card should show the correct count of assigned connections immediately after assignment, without requiring a manual page refresh.

## Actual Behavior

The card shows `0` until the page is manually refreshed.

## Root Cause

The Environment card's count comes from `getEnvironmentConnections`, an RTK Query hook tagged `providesTags: ["Environment_environments"]`.

The `addConnectionToEnvironment`/`removeConnectionFromEnvironment` mutations only invalidate `invalidatesTags: ["Connection_API_Connections"]` — a different tag — so the card's cached data never refetches after a successful assignment.

This happens because these two mutations live in the `connection` construct (`schemas/constructs/*/connection/api.yml`), and `build/bundle-openapi.js`'s `prefixTags()` function unconditionally namespaces every tag with that construct's own `info.title`. There is currently no mechanism for a tag declared in one construct file to resolve to another construct's namespace, so `connection/api.yml` cannot express "this endpoint also invalidates the `environment` construct's cache."

## Proposed Fix

Post-process the generated RTK file (`build/generate-rtk.js`) to inject the missing `Environment_environments` tag into `invalidatesTags` for these two operations, following the same pattern as the existing `guardOptionalQueryParams` post-processing step already in that file.

Faisal77666:
## Description

The Environment card's "Assigned Connections" count uses `getEnvironmentConnections`, tagged `Environment_environments`. The `addConnectionToEnvironment`/`removeConnectionFromEnvironment` mutations only invalidated `Connection_API_Connections`, so the card never refetched after a successful assignment — it kept showing a stale count even after connections were correctly assigned (confirmed via the Connections table, which uses a different, correctly-invalidated tag).

**Root cause:** these operations live in the `connection` construct (`schemas/constructs/*/connection/api.yml`), and `bundle-openapi.js`'s `prefixTags()` unconditionally namespaces tags to their own file's `info.title`. There is no OpenAPI-level way to make a `connection` endpoint invalidate an `environment`-namespaced tag across construct files.

**Fix:** post-process the generated RTK file to inject the missing `Environment_environments` tag into `invalidatesTags` for these two operations, following the same pattern as the existing `guardOptionalQueryParams` post-processing step.

## Before / After

**Before:** Environment card shows "0 Assigned Connections" after assigning a connection, requiring a manual page refresh to see the correct count.

**After:** Environment card updates immediately after assigning or removing a connection, with no manual refresh needed.

## Notes for Reviewers

This PR fixes #1044

## Signed commits

### DIFF
--- a/build/generate-rtk.js
+++ b/build/generate-rtk.js
@@ -248,6 +248,80 @@ function guardOptionalQueryParams(filePath) {
 }
 
 /**
+ * Post-process generated RTK file to add cross-construct cache invalidation
+ * for operations whose OpenAPI tags cannot express it.
+ *
+ * `addConnectionToEnvironment` / `removeConnectionFromEnvironment` live in
+ * the `connection` construct (schemas/constructs/<version>/connection/api.yml), so
+ * bundle-openapi.js's prefixTags() namespaces their tags under
+ * `Connection_API_*`. There is no mechanism for a tag declared in one
+ * construct file to resolve to another construct's namespace, so these
+ * mutations can never automatically invalidate `Environment_environments`,
+ * the tag `getEnvironmentConnections` provides -- even though assigning or
+ * removing a connection changes exactly that data. Without this, the
+ * Environment UI's "Assigned Connections" count silently goes stale after
+ * every assign/remove action.
+ *
+ * This manually injects the missing tag into `invalidatesTags` for the
+ * affected operations after codegen runs.
+ *
+ * @param {string} filePath - Absolute path to the generated TS file
+ * @returns {number} count of operations patched
+ */
+function addCrossConstructInvalidation(filePath) {
+  if (!paths.fileExists(filePath)) {
+    logger.warn(`Post-process skipped: ${filePath} not found.`);
+    return 0;
+  }
+  const content = fs.readFileSync(filePath, "utf8");
+
+  const targets = ["addConnectionToEnvironment", "removeConnectionFromEnvironment"];
+  const extraTag = "Environment_environments";
+
+  let patched = content;
+  let count = 0;
+
+  for (const opName of targets) {
+    // Find the operation's starting position, then look for invalidatesTags
+    // within a bounded window right after it. This avoids both the original
+    // cross-operation matching risk and the inner-brace false-match issue
+    // (query blocks close with their own "}),", which a naive block regex
+    // can match before reaching the operation's actual invalidatesTags line).
+    const opStart = patched.indexOf(`${opName}: build.mutation`);
+    if (opStart === -1) {
+      logger.warn(`Post-process note: ${opName} not found; skipping cross-construct tag injection.`);
+      continue;
+    }
+    const searchWindow = patched.slice(opStart, opStart + 2000);
+    const tagsMatch = /invalidatesTags:\s*\[([^\]]*)\]/.exec(searchWindow);
+    if (!tagsMatch) {
+      logger.warn(`Post-process note: ${opName} has no invalidatesTags array within range; skipping.`);
+      continue;
+    }
+    if (tagsMatch[1].includes(extraTag)) {
+      continue;
+    }
+    const existingTags = tagsMatch[1].trim();
+    const newTagsInner = existingTags.length > 0
+      ? `${existingTags}, "${extraTag}"`
+      : `"${extraTag}"`;
+    const oldFull = tagsMatch[0];
+    const newFull = `invalidatesTags: [${newTagsInner}]`;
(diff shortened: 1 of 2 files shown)


build/generate-rtk.js
```diff
@@ -247,6 +247,66 @@ function guardOptionalQueryParams(filePath) {
...
+      logger.warn(`Post-process note: ${opName} not found; skipping cross-construct tag injection.`);
+      continue;
+    }
+    if (match[2].includes(extraTag)) {
+      continue;
+    }
+    const newTags = `${match[2].trim()}, "${extraTag}"`;
+    patched = patched.replace(opRe, `$1${newTags}$3`);
+    count += 1;
+  }
```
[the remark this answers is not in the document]
Faisal77666: Thanks for catching this — updated the approach to find the operation's start position first, then search for `invalidatesTags` within a bounded window, rather than relying on a single block-matching regex. This also handles the empty-array case (`invalidatesTags: []`) cleanly, and avoids the inner-`query` closing brace being mistaken for the operation's end. Verified the generated output still correctly includes `Environment_environments` for both operations after the change.

chaitanyamedidar (commented):
@Faisal77666 Post-processing generated RTK with string/regex rewrites in `generate-rtk.js` is fragile, and it isn't required to fix this issue [here](https://github.com/meshery/meshery/issues/20628).

build/generate-rtk.js
```diff
@@ -290,6 +364,9 @@ async function generateRtkClient(rtk) {
     if (rtk.outputFile) {
       guardOptionalQueryParams(paths.fromRoot(rtk.outputFile));
     }
+    if (rtk.name === "meshery" && rtk.outputFile) {
```
winkletinkle, line 367: What's this for? Why is "meshery" special here?

build/generate-rtk.js
```diff
@@ -290,6 +364,9 @@ async function generateRtkClient(rtk) {
     if (rtk.outputFile) {
       guardOptionalQueryParams(paths.fromRoot(rtk.outputFile));
```
leecalcote, line 365: `guardOptionalQueryParams` patches around a tracked upstream codegen bug ([reduxjs/redux-toolkit#5018](https://github.com/reduxjs/redux-toolkit/issues/5018)) with a fail-loud regression guard and is _not_ a convention to follow, @Faisal77666.

leecalcote (changes requested):
@Faisal77666 thank you for chasing down this improvement. Let's close out this PR and redirect your efforts to the core Meshery repo at meshery/ui/rtk-query/connection.ts (lines ~164-174). This is the file that would naturally house a fix for addConnectionToEnvironment/removeConnectionFromEnvironment.

This file  has  for a sibling connection mutation that is a good reference - ```mesheryApi.enhanceEndpoints({ endpoints: { performConnectionAction: { invalidatesTags: [TAGS.CONNECTIONS] } } })```

In general, this approach of regex-patching the generated typescript/rtk/meshery.ts text is something we want to avoid.

Faisal77666:
> @Faisal77666 thank you for chasing down this improvement. Let's close out this PR and redirect your efforts to the core Meshery repo at meshery/ui/rtk-query/connection.ts (lines ~164-174). This is the file that would naturally house a fix for addConnectionToEnvironment/removeConnectionFromEnvironment.
> 
> This file has for a sibling connection mutation that is a good reference - `mesheryApi.enhanceEndpoints({ endpoints: { performConnectionAction: { invalidatesTags: [TAGS.CONNECTIONS] } } })`
> 
> In general, this approach of regex-patching the generated typescript/rtk/meshery.ts text is something we want to avoid.

Thanks @leecalcote  — I saw you opened #1052 building on this branch with a more robust version (delete operations, both clients, fail-loud handling, tests). Just want to make sure I'm not duplicating effort: should I hold off on the connection.ts/enhanceEndpoints approach you mentioned above, since #1052 looks like it's landing the schemas-based fix instead? Happy to go either way, just want to follow whichever direction you'd prefer going forward.

PR #1056, opened 2026-07-14
Align four cloud operations with the deployed meshery-cloud server contract

hortison:
## Summary

Schema/handler parity fixes for cloud operations whose generated RTK client calls could never succeed against the deployed meshery-cloud server (each 404'd or 400'd):

| Op | Was | Now (server truth) |
|---|---|---|
| `getKeyById` / `deleteKey` (key v1beta2) | `/api/auth/key/{keyId}` | `/api/auth/keys/{keyId}` (plural collection path; singular 404s) |
| `issueIndefiniteLifetimeToken` (token v1beta3) | required query param `owner` | `userId` (handler requires it, 400s otherwise; `owner` was never read) |
| `approveSignupRequest` / `denySignupRequest` (invitation v1beta3) | no inputs declared | the 11 query params `parseSignupRequestParams` reads (`id`, `firstName`, `lastName`, `email`, `occupation`, `organization`, `role`, `formType`, `status`, `taskId`, `taskLink`) |
| `getPattern` (design v1beta3) | no `metrics` param | optional `metrics` boolean the handler reads |

Also audited `deletePatterns` (POST `/api/content/patterns/delete`) and `unPublishCatalogContent` (POST `/api/catalog/content/{type}/unpublish`): these stay as POST sub-resources per `docs/http-api-design.md` (DELETE bodies get stripped by proxies); meshery-cloud will add those canonical routes server-side (companion PR) while keeping its legacy DELETE routes for Meshery Server's remote provider.

## Context

Unblocks the meshery-cloud RTK consumer swap (retiring the hand-authored `ui/api/api.ts` endpoints in favor of the generated `cloudApi` hooks). No working consumer exists for any of these ops today - all were broken against production - so this is a parity correction, not a wire break.

## Verification

- `make validate-schemas` - no blocking violations
- `make consumer-audit` - Consumer Only: 0 (both repos)
- `make test-rtk`, `make test-golang` - green
- Regenerated `typescript/rtk/cloud.ts` + Go models included

PR #1057, opened 2026-07-15
Integer pagination params; org domain in listings; two response shapes matched to server

hortison:
## Summary

Four parity/uniformity corrections unblocking the meshery-cloud RTK consumer swap (follow-up to #1056):

1. **Integer pagination params** - every `page`/`pageSize`/`pagesize` query param across v1beta2/v1beta3 cloud constructs is now `type: integer` (was a string/integer mix). Wire-compatible (query strings carry the same digits); generated TS args become numeric, matching every caller. Published v1alpha1/v1beta1 versions untouched.
2. **organization** - `AvailableOrganization` gains `domain` (the entity schema already had it; the listing projection did not). The deployed listing responses carry it; custom-domain org-switch logic reads it.
3. **team** - `getTeamUsers` returns `TeamMembersPage`: the deployed handler (`GetUsersOfTeam`) serves member rows with role names under `data`, not raw users-teams mappings (schema/handler parity).
4. **academy** - `getAcademyAdminSummary` response typed as `InstructorConsoleSummary` (orgConfig, learner totals, curricula/registration summaries, test outcomes) instead of bare `object`, mirroring the deployed handler struct.

## Verification

- `make consumer-audit` - Consumer Only: 0 (both repos)
- Regenerated clients inspected: `GetTeamUsersApiResponse.data`, `AvailableOrganization.domain`, typed academy summary, numeric page args

Generated artifacts intentionally not committed (automation regenerates on master).

### DIFF
--- a/schemas/constructs/v1beta2/core/api.yml
+++ b/schemas/constructs/v1beta2/core/api.yml
@@ -974,7 +974,9 @@ components:
       in: query
       description: Get responses by pagesize (pass all to get all responses)
       schema:
-        type: integer
+        # Stays string: the contract admits the literal "all" sentinel,
+        # which the server middleware expands to an uncapped page.
+        type: string
     pageSize:
       name: pageSize
       in: query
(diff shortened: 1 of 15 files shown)


hortison:
Good catch - reverted `pagesizeWithAll` to `type: string` with a comment documenting the `all` sentinel; the plain numeric params stay integer.

PR #1062, opened 2026-07-16
Define /api/system/kubernetes endpoints; add transient reachable flag to K8sContext

leecalcote:
**Notes for Reviewers**

Part of migrating meshery/meshery's hand-declared RTK endpoints onto the generated client (schemas-first mandate). This PR documents the `/api/system/kubernetes*` slice - the endpoints `ui/rtk-query/connection.ts` currently reaches with local `builder.mutation`/`builder.query` declarations.

### Endpoints added (`schemas/constructs/v1beta1/system/api.yml`, all `x-internal: ["meshery"]`)

| Operation | Route | Notes |
|---|---|---|
| `addKubernetesConfig` | `POST /api/system/kubernetes` | multipart kubeconfig import; optional `contexts` (per-context options) and `selectedContexts` (scoping) form fields; responds with `registeredContexts` / `connectedContexts` / `ignoredContexts` / `erroredContexts` buckets of `K8sContext` |
| `discoverKubernetesContexts` | `POST /api/system/kubernetes/contexts` | parse-only discovery for pre-import selection; surfaces unreachable contexts flagged via `reachable` |
| `pingKubernetes` | `GET /api/system/kubernetes/ping?connectionId=` | API-server version probe; `server_version` preserves this endpoint's published snake_case wire casing within this API version (documented on the property) |

Shapes were derived from the server handlers (`server/handlers/k8sconfig_handler.go`: `addK8SConfig`, `GetContextsFromK8SConfig`, `KubernetesPingHandler`), not from the UI's assumptions.

### K8sContext (`schemas/constructs/v1beta3/connection/api.yml`)

Adds the optional transient `reachable` flag to the `K8sContext` projection. Discovery and import responses carry it on the wire (`json:"reachable"` on the server struct) but the schema omitted it; with `additionalProperties: false` the schema would otherwise reject real payloads when reused for these endpoints.

### Deliberate choices

- The vestigial `DELETE /api/system/kubernetes` no-op handler is left undocumented.
- Generated hook names match what meshery/meshery already exports (`useAddKubernetesConfigMutation`, `useDiscoverKubernetesContextsMutation`, `useLazyPingKubernetesQuery`), so the downstream swap is a drop-in behind the existing thin wrappers.
- Generated artifacts (`models/`, `typescript/generated/`) are not committed; master automation regenerates them per AGENTS.md.

### Validation

- `make consumer-audit` - Consumer Only: 0
- `make build` (bundle, Go + RTK + TS generation, permissions, `test-golang`) - exit 0; bundle gains exactly the 3 new operations
- Generated Go: `Reachable bool json:"reachable"` matches the server struct

### Follow-up (meshery/meshery, after release + dependency bump)

Swap `addKubernetesConfig` / `discoverKubernetesContexts` / `pingKubernetes` in `ui/rtk-query/connection.ts` to the generated hooks, keeping `Connection_API_Connections` invalidation via `enhanceEndpoints` (same pattern as `performConnectionAction`).

**[Signed commits](https://github.com/meshery/schemas/blob/master/CONTRIBUTING.md)**

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -13,10 +13,11 @@ jobs:
     runs-on: ubuntu-latest   # ubuntu-24.04 is fine too, but latest is usually better
 
     steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
       - name: Label PR based on changed paths
         uses: actions/labeler@v6
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
-          # Optional: if you put config somewhere else
-          # configuration-path: .github/my-labels.yml
           # sync-labels: true     # remove labels that no longer match (careful!)
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -14,7 +14,7 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@v7
 
       - name: Label PR based on changed paths
         uses: actions/labeler@v6
--- a/schemas/constructs/v1beta1/system/api.yml
+++ b/schemas/constructs/v1beta1/system/api.yml
@@ -148,6 +148,108 @@ paths:
         "500":
           description: Email configuration validation failed or the send attempt errored.
 
+  /api/system/kubernetes:
+    post:
+      x-internal: ["meshery"]
+      tags:
+        - System
+      summary: Import a kubeconfig
+      description: >-
+        Uploads a kubeconfig and registers each discovered context as a
+        Kubernetes connection. Unreachable contexts are still registered as
+        discovered connections; reachability only gates the transition to
+        connected. The optional `contexts` and `selectedContexts` form fields
+        scope and configure the import.
+      operationId: addKubernetesConfig
+      requestBody:
+        required: true
+        content:
+          multipart/form-data:
+            schema:
+              $ref: "#/components/schemas/AddKubernetesConfigPayload"
+      responses:
+        "200":
+          description: Discovered contexts bucketed by import outcome.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/AddKubernetesConfigResponse"
+        "400":
+          description: >-
+            Missing or unparsable kubeconfig, or malformed `contexts` /
+            `selectedContexts` JSON.
+        "401":
+          $ref: "#/components/responses/401"
+        "500":
+          description: Failed to retrieve the user token from the request context.
+
+  /api/system/kubernetes/contexts:
+    post:
+      x-internal: ["meshery"]
+      tags:
+        - System
+      summary: Discover kubeconfig contexts
+      description: >-
+        Parses an uploaded kubeconfig and returns its contexts - including
+        unreachable ones, flagged via `reachable` - without persisting any
+        connection. Lets a client present the discovered contexts for
+        selection before importing them.
+      operationId: discoverKubernetesContexts
+      requestBody:
+        required: true
+        content:
+          multipart/form-data:
+            schema:
+              $ref: "#/components/schemas/DiscoverKubernetesContextsPayload"
+      responses:
+        "200":
+          description: Contexts discovered in the uploaded kubeconfig.
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: "../../v1beta3/connection/api.yml#/components/schemas/K8sContext"
+        "400":
+          description: Missing or unparsable kubeconfig.
+        "401":
+          $ref: "#/components/responses/401"
+        "500":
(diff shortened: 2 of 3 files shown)


.github/workflows/labeler.yml
```diff
@@ -13,10 +13,11 @@ jobs:
     runs-on: ubuntu-latest   # ubuntu-24.04 is fine too, but latest is usually better
 
     steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
```
marblom007:
```suggestion
        uses: actions/checkout@v7
```

schemas/constructs/v1beta3/connection/api.yml
```diff
@@ -1223,6 +1223,16 @@ components:
           x-go-type-skip-optional-pointer: true
           x-oapi-codegen-extra-tags:
             json: "connectionId,omitempty"
+        reachable:
+          type: boolean
+          description: >-
+            Whether the context's API server answered a version probe when the
+            kubeconfig was processed. Populated during kubeconfig discovery and
```
marblom007: This should not be Kubernetes specific. Ad hoc connectivity checks are performed on a broad range of different types of connections, different systems.

leecalcote: Addressed in b95c61b1. Reachability is now defined once as a shared `ConnectionReachability` schema (a point-in-time connectivity probe result, never persisted), and `K8sContext.reachable` references it, keeping only a site note about kubeconfig discovery surfacing unreachable contexts. Generated Go stays wire-identical (`Reachable bool json:"reachable"`), and `ConnectionReachability` is exported for the other probe surfaces: the upcoming telemetry slice references the same schema, since the Prometheus ping response carries the identical `reachable` field.

schemas/constructs/v1beta3/connection/api.yml
```diff
@@ -1223,6 +1223,26 @@ components:
           x-go-type-skip-optional-pointer: true
           x-oapi-codegen-extra-tags:
             json: "connectionId,omitempty"
+        reachable:
+          $ref: "#/components/schemas/ConnectionReachability"
+          description: >-
+            Whether this context's API server answered the probe run while its
```
marblom007, line 1229: This API is not specific to Kubernetes and as such should not include any Kubernetes-specific language other than by way of an example (which would be good to have). The rest of the verbiage needs to be technology agnostic.

leecalcote: Reworded in 5aab0148 - the description is now technology-agnostic (point-in-time probe of the connected system's endpoint, never persisted), with Kubernetes kept only as the single example.

schemas/constructs/v1beta1/system/api.yml
```diff
@@ -148,6 +148,108 @@ paths:
         "500":
           description: Email configuration validation failed or the send attempt errored.
 
+  /api/system/kubernetes:
```
aabidsofi19, line 151: @leecalcote  i think this is now handled from the connection post endpoint itself ( or atleast i planned to )

schemas/constructs/v1beta1/system/api.yml
```diff
@@ -148,6 +148,108 @@ paths:
...
+                items:
+                  $ref: "../../v1beta3/connection/api.yml#/components/schemas/K8sContext"
+        "400":
+          description: Missing or unparsable kubeconfig.
+        "401":
+          $ref: "#/components/responses/401"
+        "500":
+          description: Failed to retrieve the provider token for the request.
+
+  /api/system/kubernetes/ping:
```
aabidsofi19, line 220: @leecalcote  there is now a generic action endpoint on connections api ( for each connection ) which supports a ping action which can be reused for kubernetes , grafana etc

schemas/constructs/v1beta1/system/api.yml
```diff
@@ -148,6 +148,108 @@ paths:
...
+        "400":
+          description: Missing `connectionId` query parameter, or the stored kubeconfig is invalid.
+        "401":
+          $ref: "#/components/responses/401"
+        "404":
+          description: No Kubernetes context found for the given connection ID.
+        "500":
+          description: Failed to reach the cluster or fetch its server version.
+
   /api/system/sync:
```
aabidsofi19, line 253: i think this api is deprecated and needs to be get rid of

schemas/constructs/v1beta1/system/api.yml
```diff
@@ -704,3 +787,99 @@ components:
           description: Kubernetes contexts currently tracked by the Meshery server.
           items:
             $ref: "#/components/schemas/SystemSessionSyncK8sContext"
+
+    AddKubernetesConfigPayload:
```
aabidsofi19, line 791: this should also be part of connection api , the payload is the connection definition schema for each connection definition

PR #1066, opened 2026-07-16
Formalize ControllerStatusValue enum on ControllerStatus.status

hortison:
**Notes for Reviewers**

Reopens #1064 (reverted in #1065 solely to drop an AI-tool identity that had leaked into that commit's `Signed-off-by` trailer) on a clean commit. Same content, same rationale:

Prompted by an audit of meshery/meshery#20686: `ui/utils/Enum.ts` hand-rolls both `CONNECTION_STATES` and `CONTROLLER_STATES` locally instead of sourcing them from `@meshery/schemas`, which is a DRY / source-of-truth violation per this repo's own charter ("meshery/schemas becomes the permanent, authoritative source of truth" once a construct is migrated).

- `CONNECTION_STATES` already exactly duplicates the existing `ConnectionStatusValue` enum (`schemas/constructs/v1beta1/connection/api.yml`) - no schema change needed there, it just wasn't exported at a stable TS import path.
- `CONTROLLER_STATES` had no schema-side equivalent at all: `ControllerStatus.status` (`schemas/constructs/v1beta1/system/api.yml`) was a free-form `string` with only an example-value description ("e.g. DEPLOYED, NOTDEPLOYED, RUNNING, CONNECTED, UNKNOWN"), even though its own description says this REST field "Replaces the `getOperatorStatus` GraphQL query" - i.e. it's meant to formalize the existing `MesheryControllerStatus` GraphQL enum (`server/internal/graphql/schema/schema.graphql` in meshery/meshery: `DEPLOYED`, `NOTDEPLOYED`, `DEPLOYING`, `UNKOWN`, `UNDEPLOYED`, `ENABLED`, `RUNNING`, `CONNECTED`).

**What changed**

1. Added a new `ControllerStatusValue` enum schema in `schemas/constructs/v1beta1/system/api.yml`, with the same 8 values (including the published `UNKOWN` spelling, preserved as-is since it's already load-bearing in the GraphQL enum) as `MesheryControllerStatus`. Marked `x-enum-casing-exempt: true` since these are pre-existing published uppercase values, not new enum literals. Referenced it from `ControllerStatus.status` via `$ref` + `x-go-type` (not a bare `$ref` - that renames the Go field itself to the ref'd type name).
2. Exported both `ConnectionStatusValue` and `ControllerStatusValue` as consumable `v1beta1` TS types in `typescript/index.ts`, next to the existing `Connection` export, so downstream repos can type-check local status constants against the schema instead of duplicating the literal list.

This only changes the Go field's *type* (`string` → `ControllerStatusValue`, itself a `string` type) - the JSON wire shape is unchanged. `system.ControllerStatus` isn't currently constructed anywhere in meshery/meshery's server code, so blast radius there is effectively zero.

**Follow-up (tracked separately, not in this PR):** meshery/meshery#20714.

**Verification**
- [x] `make build` (bundle-openapi, generate-golang, generate-ts, generate-rtk, generate-permissions, build-ts)
- [x] `go test ./...`
- [x] `npm run build`
- [x] `make validate-schemas`
- [x] `make consumer-audit`

### DIFF
--- a/schemas/constructs/v1beta1/system/api.yml
+++ b/schemas/constructs/v1beta1/system/api.yml
@@ -408,6 +408,26 @@ components:
           minLength: 1
           maxLength: 1024
 
+    ControllerStatusValue:
+      type: string
+      description: >-
+        Current status of a single Meshery controller (operator, MeshSync, or
+        broker). Mirrors the MesheryControllerStatus GraphQL enum
+        (server/internal/graphql/schema/schema.graphql) during the ongoing
+        migration of controller-status consumers from GraphQL to this REST
+        API; the literal values (including the published "UNKOWN" spelling)
+        are load-bearing and must not be changed independently of that enum.
+      x-enum-casing-exempt: true
+      enum:
+        - DEPLOYED
+        - NOTDEPLOYED
+        - DEPLOYING
+        - UNKOWN
+        - UNDEPLOYED
+        - ENABLED
+        - RUNNING
+        - CONNECTED
+
     ControllerStatus:
       type: object
       description: >-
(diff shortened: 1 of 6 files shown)


schemas/constructs/v1beta1/system/api.yml
```diff
@@ -408,6 +408,26 @@ components:
           minLength: 1
           maxLength: 1024
 
+    ControllerStatusValue:
```
marblom007, line 411: Controller states should not be a repeat of Connection States.

marblom007 (approved):

PR #1073, opened 2026-07-18
Add registry construct defining the /api/registry API surface

marblom007:
## Description

This PR canonicalizes Meshery's capabilities-registry API surface as `/api/registry`, displacing the legacy `meshmodel` naming, per the [mailing-list proposal](https://groups.google.com/a/meshery.io/g/developers/c/Owh7n9g566M/m/93sDbiNqAQAJ) and meshery/meshery#17904 / meshery/meshery#17965.

**This PR** (schemas, source of truth first):
- **New `v1beta1/registry` construct** documenting the full registry API under `/api/registry`: model / component / relationship / category / registrant browsing, `register` (import), `generate`, `export`, `validate`, and `{entityType}/status` operations. Page envelopes use canonical camelCase (`page`, `pageSize`, `totalCount`); the snake_case keys Meshery Server still emits alongside are documented as `deprecated` for the migration window.
- **Moves `POST /api/meshmodels/register`** (`registerMeshmodels`) out of the v1beta2 model construct into the registry construct as `POST /api/registry/register` (`registerRegistryEntities`). The request content type is corrected from `multipart/form-data` to the `application/json` body the server actually decodes (`schemav1beta1.ImportRequest`). The op is now `x-internal: ["meshery"]` only - meshery-cloud's router never mounted this path (its registry listing lives under `/api/integrations/meshmodels/*`, unchanged here).
- **Renames in place** (operationIds unchanged, so generated hook names are stable):
  - `/api/meshmodels/relationships/evaluate` -> `/api/registry/relationships/evaluate` (evaluation construct)
  - `/api/meshmodels/connections` and `/api/meshmodels/connections/{connectionDefinitionId}` -> `/api/registry/connections...` (connection construct)
- **Excludes `v1beta1/registry` from Go model generation** for now: Meshery Server keeps its local response envelopes during the `/api/meshmodels` deprecation window (they dual-emit legacy snake_case pagination keys). The exclusion comment documents when to lift it.

## Coordinated changes (multi-repo)

- **meshery/meshery** (PR to follow): registers every registry handler under `/api/registry/...` as canonical and keeps `/api/meshmodels/...` + `/api/meshmodel/...` as deprecated aliases of the same handlers, so **no existing client breaks**. UI/mesheryctl/docs move to `/api/registry` and consume the hooks generated from this construct.
- **meshery-extensions**: Kanvas UI calls `/api/meshmodels/*` (hand-rolled); aliases keep it working, follow-up PR migrates it.
[...]

PR #1074, opened 2026-07-18
State the registry alias contract precisely

marblom007:
Doc-only follow-up to #1073: the registry construct description implied every operation is also aliased under the singular `/api/meshmodel` prefix. Meshery Server aliases the full surface only under `/api/meshmodels`; the singular prefix retains just the `components/register` registrant route. Matches the wording shipped in meshery/meshery#20775.

No wire or generated-client change; rides the next regular release.


leecalcote (approved):
Doc-only precision fix matching the shipped alias contract.

PR #1091, opened 2026-07-24
Generate core connection kind constants for Go and TypeScript

aabidsofi19:
Connection kinds like `kubernetes`, `grafana` and `prometheus` are currently re-declared as string literals in every consumer. This defines them once in the schema and generates the constants from it.

- Add a `CoreKind` string enum to the v1beta3 connection schema, naming the connection kinds that receive bespoke, kind-specific handling (dedicated import flows, telemetry integrations, credential forms, ping/health checks).
- The connection `kind` field itself stays an **open-ended string** — the enum is deliberately not referenced by any field, so it names the well-known kinds without closing the set. It is emitted because `skip-prune: true` keeps unreferenced component schemas.
- Go constants come for free from oapi-codegen (`connection.CoreKindKubernetes`, …). `x-enum-varnames` is required here because oapi-codegen only prefixes enum constants on collision, and would otherwise emit bare `Kubernetes` / `Grafana` into the package namespace.
- Add `build/generate-enums-ts.js` for the TypeScript side. `openapi-typescript` emits string enums as a type-only union that is erased at runtime, so there is no value to compare against; the generator emits a real `as const` object plus its derived type for any enum annotated with `x-ts-const`.
- Output goes to `typescript/constants/`, not `typescript/generated/` — `generate-schema-dts.js` copies the latter verbatim into `dist/` as `.d.ts`, which is only valid for type-only files. `typescript/index.ts` re-exports with `export *`, so a newly annotated enum flows through without editing it.
- Optional `x-ts-type` sets the TypeScript type name when the schema name is too generic un-scoped: Go names are qualified by their construct package (`connection.CoreKind`), while the TypeScript names are exported from the `@meshery/schemas` root (`CoreConnectionKinds` / `CoreConnectionKind`).
- Wire `generate-enums-ts` into `make build` and add `tests/generate-enums-ts.test.js` (8 tests), including a drift check that parses the enum out of `api.yml` and asserts the committed TypeScript and Go constants still match it.
- Document the annotation in `docs/schema-authoring-reference.md`.

### Verification

- `make validate-schemas` clean; `go build ./...` and `go test ./...` pass.
- `node --test tests/generate-enums-ts.test.js` — 8/8; RTK regression tests still 2/2.
- `npm run build` emits `CoreConnectionKinds` in `dist/index.js` and `CoreConnectionKind` in `dist/index.d.ts`.
- RTK client regeneration produces no diff (no new paths).

### Follow-up

A companion PR in `meshery/meshery` removes the hardcoded `CONNECTION_KINDS` map from `ui/utils/Enum.ts` and consumes `CoreConnectionKinds` from here. It needs a published `@meshery/schemas` containing this export before it can go green.

## Summary by CodeRabbit

- **New Features**
  - Added TypeScript runtime constants and union types for supported OpenAPI string enums.
  - Added core connection kind constants for GitHub, Grafana, Kubernetes, Meshery, and Prometheus.
  - Made generated constants available through the main TypeScript package entry point.

- **Documentation**
  - Documented the `x-ts-const` schema annotation and generated output behavior.

- **Tests**
  - Added validation for enum generation, naming, exports, and synchronization with API and Go definitions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,7 +47,7 @@ generate-site-index: site-data-generate
 #-----------------------------------------------------------------------------
 # OpenAPI spec
 #-----------------------------------------------------------------------------
-.PHONY: setup generate-ts publish-ts bundle-openapi generate-golang generate-rtk test-rtk test-ts golangci validate-schemas validate-schemas-strict audit-schemas audit-schemas-full audit-schemas-style-full audit-schemas-debt-full
+.PHONY: setup generate-ts generate-enums-ts test-enums-ts publish-ts bundle-openapi generate-golang generate-rtk test-rtk test-ts golangci validate-schemas validate-schemas-strict audit-schemas audit-schemas-full audit-schemas-style-full audit-schemas-debt-full
 
 ## (Re)Initialize Golang (go.mod) and Node (package.json) manifests
 setup:
@@ -57,6 +57,15 @@ setup:
 ## Generate typescript library, json templates, yaml templates
 generate-ts:
 	npm run generate:types
+
+## Generate TypeScript runtime constants for `x-ts-const` OpenAPI enums (requires bundle-openapi)
+generate-enums-ts:
+	npm run generate:enums:ts
+	$(MAKE) --no-print-directory test-enums-ts
+
+## Run enum constant generation regression tests
+test-enums-ts:
+	node --test tests/generate-enums-ts.test.js
 
 ## Bundle Typescript library, json templates, yaml templates
 build-ts: generate-ts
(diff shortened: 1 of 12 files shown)


sarajkrishnasingh (approved):
Reviewed as a maintainer.

Adds generated core connection-kind constants for Go and TypeScript (the single source of truth for
connection kinds), plus the generator (`build/generate-enums-ts.js`), its tests, and schema-authoring
docs. Purely additive: +677/-2, the only two deletions are a Makefile target rename; it adds new
exports (`typescript/constants/v1beta3/connection.ts`, the Go constants) and touches no existing
contract, so existing consumers are unaffected.

This unblocks meshery#20949 ("drive connection kinds from the registry instead of a hardcoded list"),
which imports these constants - without them the meshery UI build and unit tests fail, and the
schema-driven mandate forbids a local workaround. Green and mergeable clean. Approving.

PR #1093, opened 2026-07-24
fix: normalize empty credential/user UUIDs before hashing registrant identity

### LINKED ISSUE #20950: [Server] Duplicate registrant connections are created per registrant kind
### Summary

Meshery creates more than one registrant Connection per registrant kind at boot, and the duplicates are indistinguishable on the Connections page. This is pre-existing behavior, surfaced while implementing #20932.

### Evidence

Model registration builds each model's registrant from the `registrant` blob hand-authored into its `model.json`, then persists it via the content-addressed `Connection.Create` (md5 over the whole v1beta1 Connection - `schemas/models/v1beta1/connection/connection_helper.go`).

Those blobs are not uniform. Counting the shipped models:

| kind | blobs carrying `credential_id` + `user_id` | blobs omitting both |
|---|---|---|
| `artifacthub` | 208 | 927 |
| `github` | 54 | 354 |

`CredentialID` and `UserID` are `*core.Uuid`. An omitted key marshals to `null`; a present zero UUID marshals to `"00000000-0000-0000-0000-000000000000"`. Different JSON gives a different md5, so each kind lands **two** distinct rows in `connections`, both referenced by `registries.registrant_id`. A live Meshery database confirms two `artifacthub` registrant rows with 4517 and 36 references respectively.

Both variants carry the same `name`, `type` and `sub_type`, so the Connections page renders two identical rows per kind with no way to tell them apart.

### Why it is not fixed in #20932

#20932 seeds the system-owned Connections for Artifact Hub and GitHub from their registered connection definitions. It deliberately writes **one deterministic canonical registrant per kind** and leaves the others untouched, because collapsing duplicates means repointing `registries.registrant_id` and deleting rows - mutating relationships during server startup, on a defect that predates that change.

### Suggested direction

Collapse duplicate registrants per kind, outside the boot path: pick a canonical registrant, repoint every `registries.registrant_id` that references a sibling, then delete the siblings. Two things worth settling first:

1. **Where it runs.** A migration is safer than boot-time seeding - it runs once, in a transaction, with a clear failure mode, instead of on every start.
2. **Stop reintroducing it.** Even after a collapse, the next registration re-creates both variants from the same non-uniform blobs. Either normalize the `registrant` blobs in the shipped `model.json` files (and in whatever generates them), or make the registrant's content-addressed identity ignore fields that are semantically empty - `credential_id`, `user_id` - so absent and zero hash alike. The second fixes the root cause for every registrant kind, including ones added later.

Option 2 is the actual root cause: the identity hash treats "unset" and "explicitly zero" as different registrants when they are the same registrant.

### Related

- #20932 - seeds system-owned connections from registered connection definitions; found this and worked around it
- #20926 - added the Artifact Hub and GitHub connection definitions

jamieplu:
## What

Make the content-addressed identity of a registrant `Connection` ignore `credential_id` and `user_id` when they are **semantically empty**, so an absent field and a present-but-zero UUID hash identically.

Fixes meshery/meshery#20950.

## Root cause

- **Symptom** - Two indistinguishable `connections` rows exist per registrant kind (confirmed live: two `artifacthub` registrant rows, 4517 vs 36 references), and any one-time collapse gets reintroduced on the next registration.
- **Root cause** - Registrant identity is `md5` over the whole v1beta1 `Connection`. `CredentialID` and `UserID` are `*core.Uuid`. A model's `registrant` blob that **omits** them marshals to `null`; one that carries a **present-but-zero** UUID marshals to `"00000000-0000-0000-0000-000000000000"`. Same registrant, two JSON encodings, two md5s -> two distinct content-addressed ids. The shipped `model.json` blobs are non-uniform in this respect, so every future registration re-creates both variants.
- **Fix** - In `models/v1beta1/connection/connection_helper.go`, `GenerateID()` now normalizes a **copy** of the `Connection` before hashing: a `CredentialID`/`UserID` that is `nil` **OR** the zero UUID (`uuid.Nil`) collapses to `nil`. Absent and zero now hash identically; a **non-zero** value still participates in identity unchanged. Normalization operates on a copy, so the caller's `Connection` and the persisted row's actual field values are untouched - only the identity computation sees the canonical form.
- **Watch for** - `CredentialID` and `UserID` are the only `*core.Uuid` pointer fields on `Connection` carrying the absent-vs-zero ambiguity, so they are the only two normalized. `ID` is a non-pointer `core.Uuid` (no null/zero ambiguity in marshalling) and is intentionally not touched.

This PR stops the **reintroduction** of duplicate rows at the source. It does **not** perform the separate one-time collapse/migration of already-duplicated rows (repoint `registries.registrant_id` to the surviving connection and delete siblings) - that remains needed and is out of scope here.

## Tests

Added `connection_helper_test.go` in the connection package proving:

- `CredentialID = nil` and `CredentialID = zeroUUID` produce the **same** id (same for `UserID`, and both empty together).
- A **non-zero** `CredentialID`/`UserID` produces a **different** id from the empty case, and two distinct non-zero values do not collide.
- No-regression baseline: two fully-identical Connections still hash equal.
- Normalization does not mutate the caller's `Connection` (zero-UUID pointers survive the call).

`go test ./models/v1beta1/connection/...` -> `ok`. `go vet` and `go build ./models/...` clean.

## Codegen

None. `GenerateID` is hand-written Go in `connection_helper.go` (not OpenAPI-derived); no generated Go/TS artifact regenerates from it.

## Downstream

`meshery` consumes `@meshery/schemas`. Once this publishes, **meshery must bump `@meshery/schemas`** to pick up the collapsed identity - until then it keeps generating both variants.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate connection IDs when credential or user identities are empty, whether represented as missing or zero-valued identifiers.
  * Preserved connection data while generating IDs.

* **Tests**
  * Added coverage for identity-neutral values, non-empty identities, unchanged input data, and consistent IDs for identical connections.

### DIFF
--- a/models/v1beta1/connection/connection_helper.go
+++ b/models/v1beta1/connection/connection_helper.go
@@ -16,13 +16,36 @@ import (
 var connectionCreation sync.Mutex //Each entity will perform a check and if the connection already doesn't exist, it will create a connection. This lock will make sure that there are no race conditions.
 
 func (h *Connection) GenerateID() (uuid.UUID, error) {
-	byt, err := json.Marshal(h)
+	// Normalize a copy before hashing so that a semantically-empty identity
+	// field (nil OR the zero UUID) collapses to a single canonical form. A
+	// model's registrant blob may omit credential_id/user_id (marshals to
+	// null) or carry a present-but-zero UUID (marshals to the all-zero
+	// string); both mean "no credential/user" yet previously produced
+	// different md5s, and therefore duplicate content-addressed connection
+	// rows for the same registrant. Normalization must not mutate the
+	// caller's Connection or the persisted row, so we operate on a copy and
+	// only the identity computation observes the canonical form.
+	normalized := *h
+	normalized.CredentialID = canonicalizeUUIDPtr(normalized.CredentialID)
+	normalized.UserID = canonicalizeUUIDPtr(normalized.UserID)
+
+	byt, err := json.Marshal(&normalized)
 	if err != nil {
 		return uuid.UUID{}, err
 	}
 
 	hash := md5.Sum(byt)
 	return uuid.UUID(hash), nil
+}
+
+// canonicalizeUUIDPtr reduces a semantically-empty UUID pointer to nil so that
+// an absent field and a present-but-zero UUID hash identically. A non-zero
+// UUID is returned unchanged so it still participates in identity.
+func canonicalizeUUIDPtr(id *core.Uuid) *core.Uuid {
+	if id == nil || *id == uuid.Nil {
+		return nil
+	}
+	return id
 }
 
 func (h *Connection) Create(db *database.Handler) (uuid.UUID, error) {
--- a/models/v1beta1/connection/connection_helper_test.go
+++ b/models/v1beta1/connection/connection_helper_test.go
@@ -1,0 +1,126 @@
+package connection
+
+import (
+	"testing"
+
+	"github.com/gofrs/uuid"
+	"github.com/meshery/schemas/models/core"
+	"github.com/stretchr/testify/assert"
+)
+
+// baseConnection returns a minimal, deterministic registrant-style Connection
+// used as the fixed baseline for identity tests.
+func baseConnection() *Connection {
+	return &Connection{
+		Name:    "artifacthub",
+		Type:    "registry",
+		SubType: "",
+		Kind:    "artifacthub",
+		Status:  ConnectionStatusRegistered,
+	}
+}
+
+func uuidPtr(u core.Uuid) *core.Uuid {
+	return &u
+}
+
+// TestGenerateID_EmptyCredentialAndUserAreIdentityNeutral asserts the core fix
+// for meshery/meshery#20950: an absent (nil) identity field and a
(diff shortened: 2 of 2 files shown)


fitzergerald (approved):
Reviewed as a maintainer.

Implements Option 2 from the #20956 review (follow-up issue #20950): the registrant's content-addressed identity now normalizes semantically-empty `credential_id` / `user_id` before hashing, so an absent field (marshals to null) and a present zero UUID (00000000-...) produce the same md5. This is the root-cause fix for the two-registrant-rows-per-kind duplication - it stops the reintroduction at the source (the shipped model.json blobs are non-uniform on these fields).

Verified:
- A non-zero credential_id/user_id still participates in identity unchanged; only the nil-or-zero case is collapsed. Normalization is on a copy - the persisted row's real field values are untouched.
- Regression test proves nil and zero-UUID hash equal (and both-empty together), while a non-zero UUID hashes differently, and two identical connections still hash equal.
- Correctly flags the downstream consumer (meshery must bump @meshery/schemas after publish) and that the one-time collapse/migration of already-duplicated rows remains a separate follow-up.

Mergeable state clean; checks passing. Approving.

Issue #735, opened 2026-04-08
Add descriptions to all operations and parameters across constructs


arjunmehta-git:
## Context

Many operations and parameters in the schema constructs lack \`summary\`, \`description\`, or both.

## Problem

- Operations without summaries produce blank entries in API documentation
- Parameters without descriptions are opaque to API consumers and generated code users
- This is a documentation completeness gap

## Proposal

Audit all \`api.yml\` files and add:
1. \`summary\` (one-line) and \`description\` (detailed) to every operation
2. \`description\` to every path, query, and header parameter
3. Pair this with the new validator rules (#718) to enforce going forward

## Scope

Touches every construct with endpoints (~20 \`api.yml\` files). Large but mechanical — each operation needs a sentence or two.

## References

Identified in schema validator audit — Section 5 items #4 and #5.

Akshay000000:
Hi! I'd love to help add descriptions to operations and parameters. I've been studying the Meshery schema structure as part of my AI Adapter contributions. Can I be assigned a subset to start?

kanishksingh23:
Hey @miacycle, @arjunmehta-git, I would like to work on this issue. Can you please assign this to me?

shteypandey28-hue:
can I work on this .. @rishiraj38

PR #1100, opened 2026-07-27
[validation] Add operation and parameter description rules (48-49)

### LINKED ISSUE #718: Validator: add operation and parameter description checks (Rules 45-46)
## Context

Rule 36 checks that schema *properties* have descriptions, but there is no equivalent check for operations or parameters.

## Problem

- Operations without \`summary\` or \`description\` produce blank entries in API documentation
- Parameters without \`description\` are opaque in generated docs and Swagger UI

## Proposal

Add two new advisory rules:
1. **Rule 45 (operation description):** Every operation should have \`summary\` or \`description\`
2. **Rule 46 (parameter description):** Every declared parameter (path, query, header) should have \`description\`

Both should be advisory (not blocking) to allow incremental adoption.

## Impact

Improves API documentation completeness. Touches existing constructs that lack descriptions.

## References

Identified in schema validator audit — Section 2a.

anurag629:
**Notes for Reviewers**

This PR fixes #718

Adds two advisory validator rules:

- Rule 48: every operation must have a summary or a description.
- Rule 49: every path, query, and header parameter must have a description.

The issue proposed rules 45 and 46, but those are taken now. 45 is partial casing migrations, 46 is sibling-endpoint parity, and 47 is in use too. So these land as 48 and 49, the next free numbers.

Both rules are advisory by default and blocking under `--strict`, same as the other design rules.

For rule 49, most parameters in these schemas are `$ref`s to `components.parameters`. So I check the shared parameters at their definition, plus any inline parameters where they are declared. A `$ref`'d usage is not checked again at the use site, since its description lives at the definition. This is the same approach rule 37 uses for `$ref`'d properties. Cookie parameters are out of scope.

How I tested it:

- `go test ./...` passes. 15 new tests cover summary/description present and absent, whitespace-only, strict severity, `$ref` skipping, cookie exclusion, and nil docs.
- `make validate-schemas` reports no blocking violations.
- Running the validator with `--warn` over the current schemas shows the new findings as advisory only, about 21 undocumented parameters. Nothing in the existing schemas is forced to change and CI is not affected. Those can be fixed separately in #735.
- `gofmt`, `go vet`, and `go build ./...` are clean. No schemas or generated artifacts were changed.

Documented in `docs/schema-authoring-reference.md` next to the per-property rules.

## Summary by CodeRabbit

* **New Features**
  * Added validation checks requiring each API operation to include a summary or description.
  * Added validation checks requiring descriptions for path, query, and header parameters.
  * Strict validation mode reports these documentation issues as blocking violations.
* **Documentation**
  * Added guidance and YAML examples for documenting operations and parameters.
* **Tests**
  * Added coverage for inline, referenced, and excluded parameters, whitespace-only content, and strict-mode behavior.


anurag629:
@ishwar170695 gentle nudge on this one. All checks are green (CodeRabbit, DCO, triage) and CodeRabbit had no change requests. Happy to adjust anything if you'd like.

ishwar170695 (commented):
Found one issue with Rule 49 around external `$ref`s.

Currently both checks skip anything with `ref.Ref != ""`:

```go
if ref == nil || ref.Value == nil || ref.Ref != "" {
    continue
}
```

The assumption seems to be that the referenced parameter will be checked at its definition, but `walkValidatedConstructSpecs` only audits the latest version of each construct. We have existing refs to older versions, e.g. `v1beta2/team/api.yml` references parameters from `v1alpha1/core/api.yml`.

So in that case the usage gets skipped here, and the referenced older `core` schema isn't audited either. An undocumented parameter can therefore bypass Rule 49 completely.

Since `kin-openapi` resolves the ref and `ref.Value` is available, can we validate the resolved parameter for external refs while still avoiding duplicate checks for local refs?

Would also add tests for an external `$ref` with and without a description.

Other than this, the Rule 48/49 implementation and tests look good.

ishwar170695:
and sorry for late review T-T

PR #1103, opened 2026-07-28
Consolidate agent skills into .agents/skills as single source of truth

jamieplu:
## What

Consolidates this repo's agent skills into a single source of truth at `.agents/skills`, and makes `.claude/skills` a relative symlink to it.

**The direction matters.** Both directories existed here as real tracked directories with *different* contents, and `.claude/skills` was the richer one. Deleting it and symlinking would have destroyed 11 tracked files. Content moves **from `.claude/skills` into `.agents/skills`**, and only then does `.claude/skills` become a link.

## File counts - before / after

| Path | Before | After |
|---|---:|---:|
| `.agents/skills` (tracked files) | 4 | **15** |
| `.claude/skills` (tracked files) | 11 | 1 (the symlink) |
| **Total skill files** | **15** | **15** |

Nothing was dropped. 4 + 11 = 15.

### Skills before

- `.agents/skills`: `chrome-devtools-axi`, `gh-axi`, `lavish`, `quota-axi`
- `.claude/skills`: `create-openapi-schemas-from-golang-models`, `cut-release`, `iterate-pr`, `openapi-schema-best-practices`

**No name collisions** - the two sets were fully disjoint, so no conflict resolution was needed and none was performed.

### Skills after (all under `.agents/skills`, reachable via both paths)

`chrome-devtools-axi`, `create-openapi-schemas-from-golang-models`, `cut-release`, `gh-axi`, `iterate-pr`, `lavish`, `openapi-schema-best-practices`, `quota-axi`

`ls .claude/skills/` and `ls .agents/skills/` produce identical listings.

## How

- Every move used `git mv`, so the diff reviews as renames (`R`) and history is preserved.
- **No skill content was edited.** This is a relocation only.
- File modes preserved (the two executable `iterate-pr` scripts stayed `100755`).

### Relative paths - checked, none broken

- Skill directories moved from `.claude/skills/<name>` to `.agents/skills/<name>` - the same depth from the repo root, so any repo-relative path is unaffected.
- The `../` occurrences inside the moved skills are OpenAPI `$ref` examples (`../../v1alpha1/core/api.yml#/...`) describing *schema* file layout in `schemas/constructs/`. They are documentation content, not skill-relative filesystem paths, and are unaffected.
- `iterate-pr/SKILL.md` already referenced its scripts as `${HOME}/.agents/skills/iterate-pr/scripts/...` - a home-directory path, unaffected by a repo-local move. Left as-is.

**No path fixes were required.**

## `.codex` / `.opencode` - verified, and deliberately NOT created

I checked each tool's actual behavior rather than assuming. Neither link is created, because **both tools already read `.agents/skills` natively at the repo level**. A link would be redundant, and would risk the same skill being discovered twice under two roots.

**OpenCode** - [official docs, `opencode.ai/docs/skills`](https://opencode.ai/docs/skills/) list six discovery locations, including project-level `.agents/skills/<name>/SKILL.md` and `.claude/skills/<name>/SKILL.md` alongside its own `.opencode/skills`.

**Codex** - confirmed in Codex's own source, `codex-rs/core-skills/src/loader.rs`:

```rust
const AGENTS_DIR_NAME: &str = ".agents";
const SKILLS_DIR_NAME:  &str = "skills";
```

`repo_agents_skill_roots()` walks from the project root to the cwd and probes `<dir>/.agents/skills` for each, registering hits with `scope: SkillScope::Repo`. There is no `.claude` constant in the loader at all. Codex also follows symlinks for repo-scope roots (`DirectorySymlinkPolicy::Follow`), so the `.claude/skills` symlink is safe either way.

Worth noting the evidence is not entirely tidy: [openai/codex#21907](https://github.com/openai/codex/issues/21907) ("Support project-level skills discovery from repository workspace") is still open and claims project-level discovery does not work. The source above and openai/codex's own checked-in `.codex/skills/` directory both contradict it, so I trusted the code over the issue.

## Also in this PR

Two small changes tied directly to the relocation:

1. `.gitignore` - `!.claude/skills/` lost its trailing slash. A trailing-slash pattern matches directories only; `.claude/skills` is now a symlink. The line is a no-op today (nothing broadly ignores `.claude/`), but it would silently stop working if a broader rule were ever added.
2. `AGENTS.md` - a short "Where skills live" note so the next person edits `.agents/skills/` and does not recreate `.claude/skills` as a real directory. The `## Maintaining this file` self-governance section was added in the same pass, as it was missing.

## Out of scope, noted not changed

`iterate-pr/scripts/reply_to_thread.py` is mode `100644` while its two sibling scripts are `100755`. Pre-existing, unrelated to this move, and harmless because `SKILL.md` invokes all three via `uv run <path>` rather than executing them directly. Left alone to keep this PR layout-only.

## Verification

```
$ diff <(ls .claude/skills/) <(ls .agents/skills/) && echo IDENTICAL
IDENTICAL

$ git ls-files .agents/skills | wc -l
15
```

## Summary by CodeRabbit

* **New Features**
  * Added comprehensive Academy API definitions covering curricula, registrations, progress tracking, quizzes, test sessions, administration, and certificates.
  * Added support for storing and retrieving quiz data and evaluation results.

* **Documentation**
  * Added OpenAPI schema best-practice guidance and Academy schema examples.
  * Added release publishing and pull request workflow guidance for contributors.

* **Chores**
  * Improved skills management and repository configuration for consistent tooling.


chellej (approved):
Reviewed as a maintainer.

- Change: 14 file(s), +16/-1, base master
- Checks at review time: 2 success

Reviewed the diff and the check results. Approving.

PR #1108, opened 2026-07-29
ci(issue-template): create Website issue template

### LINKED ISSUE #1098: [CI] Create a "Website" GitHub issue template
### Current Behavior
https://schemas.meshery.io doesn't have a dedicated issue template.

### Desired Behavior
Copy the other template materials, adjusting to be specific to the website in this repo.

---
- 🎨 Wireframes and designs for Meshery UI in [Figma](https://www.figma.com/file/SMP3zxOjZztdOLtgN4dS2W/Meshery-UI)
- 🧪 [Meshery Test Plan Spreadsheet](https://docs.google.com/spreadsheets/d/13Ir4gfaKoAX9r8qYjAFFl_U9ntke4X5ndREY1T7bnVs/edit#gid=0)

AnkitRewar11:
## Notes for Reviewers
This PR fixes #1098

This PR creates a dedicated "Website" GitHub issue template (`.github/ISSUE_TEMPLATE/website.md`) for `schemas.meshery.io`, following the repository's casing rules and issue template conventions.

### Changes Included
- Added `.github/ISSUE_TEMPLATE/website.md` with `[Website]` title prefix and `area/website`, `framework/jekyll` labels.
- Included standard contributor guides and handbook footer links as specified in #1098.

## Signed commits

## Summary by CodeRabbit

* **Documentation**
  * Added a dedicated issue template for reporting website-related problems.
  * The template guides reporters to describe current and desired behavior and include optional screenshots or logs.
  * Added links to relevant contributor resources.


Maanvi212006 (commented):
Went through this end to end (diff, content, labels config, and checked live labels on GitHub) — LGTM overall.

One thing to fix before merge: the template uses `area/website` as a label, but that label doesn't actually exist in the repo yet. Right now it'll just silently not get applied when someone files an issue. Could we create it first, or use an existing label instead?

AnkitRewar11:
@Maanvi212006 Thanks for catching that! If we have to use a pre-existing label here, I think `area/docs` is the best option as it already serves `schemas.meshery.io`. Do you think that makes sense, or would you prefer introducing a dedicated `area/website` label instead?

Maanvi212006:
@AnkitRewar11 maybe we can introduce a new label dedicated to website. 
@rishiraj38 @ishwar170695 suggestions?

ishwar170695:
@Maanvi212006  I lean towards using `area/docs` for now. 
Looking through the repository, I don't see evidence that the website is treated as a separate area, and we already have `framework/jekyll` to distinguish website-related issues. Unless there's a broader plan to split website work into its own category, reusing `area/docs` seems sufficient.

AnkitRewar11:
Sounds good! I'll update the label to `area/docs` and push the changes. Since it's already being used for both documentation and website-related content, I think it's the most appropriate choice for now.
@Maanvi212006 @ishwar170695

AnkitRewar11:
@ishwar170695 Hello, good morning! Please, whenever you have time, take a look at this. Have a nice day! 🙌

Maanvi212006 (approved):
LGTM!

ishwar170695 (approved):

.github/ISSUE_TEMPLATE/website.md
```diff
@@ -0,0 +1,26 @@
+---
+name: 🌐 Website issue
+about: Issues related to website.
+title: '[Website]'
+labels: 'area/docs, framework/jekyll'
```
winkletinkle, line 5: Docs?

Perhaps, a new dedicated label is needed.

AnkitRewar11: Sir, I had initially suggested using `area/website`, as you can see in the earlier conversation above. However, I later changed it to `area/docs`. Although `area/website` was a new label, after our discussion I felt that `area/docs` was the most appropriate choice, so I updated it accordingly.

emphor11:
LGTM!

AnkitRewar11:
Hey, could you please have a look when you get some time and let me know if there are any other changes needed? Otherwise, I think everything looks good. @banana-three-join

banana-three-join (changes requested):
I don't agree with the `area/docs` label. There's currently a `docs` folder and a `documentation.md` template which already uses the `'area/docs, framework/jekyll, language/markdown'` labels so there needs to be a bigger differentiator between the two of them to have a better track of the issues. 

You can use something like `area/site` since that's the folder where the main logic for `https://schemas.meshery.io/` resides.

AnkitRewar11:
I also agree with your point about using the Area/Site label. However, @ishwar170695 and @Maanvi212006 suggested that since this is a new label, using Area/Docs would be more appropriate. That said, we can go with Area/Site if you think that’s better. I’ll push a new commit with the Area/Site label. @banana-three-join

banana-three-join:
I did see both of their points and while I do think they're right, I also think that applying a placeholder label in the meanwhile because there isn't another label available, isn't the preferred approach. Let me go take a look at to how the other repos do it so I can comeback with a more conclusive answer.

Thanks to both of you for helping out over in this area! @Maanvi212006 @AnkitRewar11

vedant21-ctr (changes requested):
A couple of final points :

The template headings are still using ### directly after the front matter. These should be ## for the top-level sections to satisfy MD001.
Regarding the label, I checked the wider Meshery org and area/website is already used for website-related issues in meshery/meshery.io. Since area/docs is already used for the documentation workflow in this repo, using area/website here would give the template a distinct category while keeping the naming consistent across the org.

I think we can settle on area/website and create the label in this repo rather than using area/docs or introducing area/site.

PR #1094, opened 2026-07-25
Add missing x-superseded-by annotations

vanshika921vd:
**Notes for Reviewers**

## Summary

This PR adds missing `x-superseded-by` annotations to deprecated schemas where newer schema versions already exist.

## Changes

- Added `x-superseded-by: v1beta2` to deprecated `v1beta1` schemas.
- Added `x-superseded-by: v1beta3` to deprecated `v1beta2` schemas.

This PR fixes #995

**[Signed commits](https://github.com/meshery/meshery/blob/master/CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin)**
- [x] Yes, I signed my commits.
 

## Summary by CodeRabbit

* **Documentation**
  * Updated API metadata to clearly identify superseded versions.
  * v1beta1 Badge, Credential, Key, Keychain, and View specifications now point to v1beta2.
  * v1beta2 Event, Invitation, Plan, and Subscription specifications now point to v1beta3.
  * Existing deprecation indicators remain unchanged.

PR #1125, opened 2026-08-04
feat(consumer-audit): opt-in superseded-construct report

### LINKED ISSUE #1107: Extend consumer-audit with an opt-in superseded-construct report
### Current Behavior

`x-superseded-by` is now carried by **33 constructs** across `v1beta1` and `v1beta2`, but **no tooling consumes it.** Outside the schema YAML its only occurrences are in generated TypeScript that passes the annotation through verbatim - `grep -rn superseded cmd/ build/` returns nothing. `validate-schemas` does not read it, and neither does `consumer-audit`.

The practical effect is that answering "is any consumer still on a superseded construct version?" is a fully manual exercise. It was done by hand for #1094's nine constructs (see `docs/superseded-construct-consumer-verification.md`), and that took a non-obvious method, because the naive check silently returns a false clean.

### Desired Behavior

Extend `cmd/consumer-audit` with an **opt-in** superseded-construct report. It already accepts exactly the inputs required - `--meshery-repo`, `--cloud-repo`, `--extensions-repo` and their `-ui` variants - so this is an additional mode rather than new plumbing.

Proposed shape:

```
make consumer-audit SUPERSEDED_REPORT=1
```

Reads `x-superseded-by` from each construct's `api.yml`, then for each consumer repo reports which version of each superseded construct it resolves, and exits non-zero only when opted in to enforcement.

Opt-in matters: downstream repos legitimately pin superseded versions, and the Phase 4.A non-deletion policy in `docs/schema-tooling.md` keeps them served indefinitely. This should be an awareness tool by default, not a gate.

### Why this is more than a grep

Three surfaces need different predicates, and only the first is a name match:

1. **Go** - the import path is version-qualified, so a direct grep works.
2. **npm `constructs/*` deep import** - also version-qualified. The published package ships every version, superseded included (`dist/constructs/v1beta1/badge/Badge.d.ts`), so this path *can* reach a superseded schema.
3. **npm bundled clients (`cloudApi` / `mesheryApi`)** - version is **not in the import at all.** Successive versions declare identical `operationId`s on near-identical paths (`v1beta1/view` and `v1beta2/view` both declare all six view operations), and `build/bundle-openapi.js` merges every version into one spec before client generation. One version wins each collision; the loser is absent from the client entirely.

So surface 3 cannot be audited from consumer code at all - it is a property of this repo, resolvable only via a discriminator (path-param rename, casing migration, successor-only operation). A tool living here is the right place for that, since it can compute the merge winner directly from the bundle instead of inferring it.

This is also the failure mode worth guarding against: auditing only surfaces 1 and 3 and concluding an npm-only consumer is clean skips surface 2, which is the one that can actually find a superseded use.

### Additional Context

Method, per-surface predicates, and the 2026-07-28 result: `docs/superseded-construct-consumer-verification.md` (added in #1105).

Follows from #1094, which added the missing annotations. Requested by @leecalcote in https://github.com/meshery/schemas/pull/1105#issuecomment-5110308922.

Himanshujha7:
**Notes for Reviewers**
## What this does

`x-superseded-by` is carried by 33 construct versions but nothing read it, so
answering "is any consumer still on a superseded construct version?" has been a
manual exercise. #1105 documented the method and worked through the nine
constructs from #1094 by hand. This PR turns that method into a tool and widens
it to all 33.

This PR fixes #1107 

Upon running:

`make consumer-audit MESHERY_REPO=../meshery SUPERSEDED_REPORT=1
`

Output:

```
Consumer Audit Report

Metric                         Schema  Meshery  Cloud  
Total endpoints                248     237      0      
Spec applies to consumer       -       120      198    
Spec targets Meshery only      50      -        -      
Spec targets Cloud only        128     -        -      
Spec targets both consumers    70      -        -      
Spec passes validation         248     -        -      
Spec only (no handlers)        192     -        -      
Spec without consumer handler  -       64       -      
Handler only (no spec)         -       177      0      

Note:
Spec only (no handlers) --> no matching handler in any audited consumer.
Spec without consumer handler --> this consumer is missing it (may exist elsewhere).

Action Needed

Meshery Server has 1 active endpoint matching the schema but is marked Cloud-only.
Audited 237 Meshery Server endpoints; 120 spec-defined endpoints apply to Meshery Server.
Out of those 120 applicable spec-defined endpoints, 3 are schema-driven.
Out of 120 spec-defined Meshery Server endpoints, 64 are unimplemented.

Note:
Out of 120 spec-defined Meshery Server endpoints, 0 are blocking schema-audit violations. Run `make audit-schemas` for details.

Superseded Construct Report

33 of 78 construct versions carry x-superseded-by; 33 carry x-deprecated.
Superseded versions stay served indefinitely (Phase 4.A non-deletion policy),
so a pinned superseded version is a migration signal, not an error.

Bundled clients (npm-bundled): no superseded construct reaches the merged spec,
so no bundled client can expose one. Computed from the build's own merge
exclusions, not inferred from consumer code.

  meshery (../meshery) - 1868 files scanned
Construct     Surface  Resolves  Files  Status                             
academy       -        -         -      not used                           
badge         -        -         -      not used                           
catalog       -        -         -      not used                           
component     go       v1beta1   19     SUPERSEDED -> v1beta2/component    
component     go       v1beta2   16     current                            
component     go       v1beta3   20     current                            
connection    go       v1beta1   12     SUPERSEDED -> v1beta3/connection   
connection    go       v1beta3   8      current                            
core          go       v1beta2   3      current                            
credential    go       v1beta2   1      current                            
design        go       v1beta1   30     SUPERSEDED -> v1beta3/design       
design        go       v1beta3   17     current                            
environment   go       v1beta1   11     SUPERSEDED -> v1beta3/environment  
environment   go       v1beta3   1      current                            
event         -        -         -      not used                           
invitation    -        -         -      not used                           
key           go       v1beta2   1      current                            
keychain      -        -         -      not used                           
model         go       v1beta1   21     SUPERSEDED -> v1beta2/model        
organization  go       v1beta2   4      current                            
plan          -        -         -      not used                           
relationship  go       v1alpha3  11     current                            
relationship  go       v1beta2   30     current                            
role          -        -         -      not used                           
schedule      go       v1beta2   1      current                            
subscription  -        -         -      not used                           
team          -        -         -      not used                           
token         -        -         -      not used                           
user          go       v1beta3   1      current                            
view          go       v1beta2   1      current                            
workspace     go       v1beta1   3      SUPERSEDED -> v1beta3/workspace    
workspace     go       v1beta3   9      current                            
    23 bundled-client imports (cloudApi/mesheryApi); version not resolvable from consumer code.
    6 superseded resolutions; 11 not used.
```

 Kindly let me know if there is a need of further checks and corrections in the report generated.
cc @marblom007 

## Summary by CodeRabbit

* **New Features**
  * Added opt-in auditing for superseded and deprecated constructs across Go, npm, and bundled client imports.
  * Added reports for successor resolution, reachability, unused constructs, dangling references, parse warnings, and annotation issues.
  * Added optional enforcement with distinct outcomes for incomplete scans, superseded usage, unresolved references, and audit failures.

* **Bug Fixes**
  * Improved construct discovery with deterministic traversal, extension filtering, directory exclusions, and scan-defect reporting.
  * Added handling for successor chains, cycles, dangling references, and deprecated versions.


ayanasarkar (commented):
Nice work turning the manual audit into a real tool 👍 Report's clear and useful.

Only thing 
CodeRabbit's right about those unchecked `Fprintln` errors around line 117-118, might trip lint. Otherwise LGTM, checks are green ✅

Himanshujha7:
Thank you for the review, @ayanasarkar 
Yes, definitely, I will try to map it to a similar pattern that the existing lint configuration accepts - will need to check that once.

cmd/consumer-audit/main.go
```diff
@@ -87,6 +91,34 @@ func main() {
...
+			fmt.Fprintf(os.Stderr, "consumer-audit: superseded report: %v\n", err)
+			os.Exit(1)
+		}
+		printSupersededReport(out, sup)
+
+		if *supersededEnforce && sup.HasSupersededUsage() {
+			fmt.Fprintln(out)
+			fmt.Fprintln(out, "superseded-enforce: at least one consumer resolves a superseded construct version.")
+			os.Exit(2)
+		}
```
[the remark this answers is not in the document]
Himanshujha7: Hi, Thanks for the review.
errcheck may well flag these, but the two lines aren't distinctive since there are multiple
bare `fmt.Fprint*(out, ...)` calls in this file, as I can see, 34 of them pre-existing with `printAuditReport` and `printTSFindings` having the identical `out io.Writer` shape.

Himanshujha7:
Hi @marblom007  @pontusringblom Kindly review this PR and let me know if it's GTG.

Maanvi212006 (commented):
-  Built clean, ran full `validation` test suite — all pass (1 expected skip)
- Manually ran the tool (`SUPERSEDED_REPORT=1`, `SUPERSEDED_ENFORCE=1`, and default) against a local meshery checkout — report, enforce exit code, and opt-in behavior all work as described

Just one thing: lint confirms the unchecked `Fprintln`/`Fprintf` errors already flagged by CodeRabbit/@ayanasarkar are still unresolved

Himanshujha7:
Hi @Maanvi212006  @ayanasarkar 
Thank you for the great review.
The `Fprintf` error flagged by CodeRabbit appears to be a pre-existing call in the file since it already appears in multiple other places in the codebase, which are untouched by me.
I have also responded to CodeRabbit the same as well.

validation/audit.go
```diff
@@ -27,12 +27,15 @@ type constructSpec struct {
 	LoadErr      error
 }
 
-// walkValidatedConstructSpecs is the single, canonical walker for the
-// schemas/constructs tree. It visits every non-deprecated construct spec in
-// deterministic sorted order and invokes fn for each one. Both the schema
-// validator (Audit) and the consumer-audit endpoint indexer use this function
-// so the walk logic, filtering, and load behaviour stay in sync.
-func walkValidatedConstructSpecs(rootDir string, fn func(constructSpec) error) error {
+// walkAllConstructSpecs is the raw traversal of the schemas/constructs tree.
```
emphor11, line 30: Could we avoid treating an incomplete construct index as a clean audit result here?

`walkAllConstructSpecs` can return successfully after skipping construct/version entries whose `api.yml` cannot be loaded, and `LoadConstructIndex` then omits those entries from `ByKey`/`ByExposed`. Because `--superseded-enforce` ultimately relies on `HasSupersededUsage()`, a missing or unreadable superseded schema can therefore disappear from the audit and produce a false-clean result.

Could we propagate the load/traversal failure, or carry an explicit incomplete-index state and make enforcement fail when the index is incomplete?

validation/source_tree.go
```diff
@@ -73,6 +120,60 @@ func (t localTree) Walk(dir string, fn func(path string) error) error {
...
+		}
+		return err
+	}
+	if !info.IsDir() {
+		return fmt.Errorf("source_tree: %q is not a directory", dir)
+	}
+
+	var paths []string
+	err = filepath.WalkDir(abs, func(p string, d fs.DirEntry, walkErr error) error {
+		if walkErr != nil {
```
emphor11, line 169: Could the audit distinguish an unscanned consumer from a consumer with no superseded usage?

`WalkFiltered` treats a missing consumer directory as a successful empty scan, and unreadable files/subtrees are skipped without recording that the traversal was incomplete. The superseded report can therefore emit `not used` for a consumer that was never fully scanned, while `--superseded-enforce` can still treat the result as clean.

Since this command is also used in enforcement mode, could we surface incomplete scans explicitly and make enforcement fail when a requested consumer cannot be fully scanned?

validation/superseded_test.go
emphor11, line 1: Could the Go import resolution account for the generator's unversioned package mappings?

The scanner currently matches only `models/<version>/<package>`, but `build/generate-golang.js` explicitly maps `v1beta1/core` to `github.com/meshery/schemas/models/core` (and the same mapping exists for `v1alpha1/core`). That published import path is therefore invisible to the current regex/index lookup.

Even though we don't currently see a Meshery consumer using this path, the audit should cover the actual generated import surface rather than silently treating supported unversioned mappings as unused. Could we derive these mappings from the generator configuration or otherwise report them as ambiguous/unresolved?

emphor11, line 1: One concern with the source scanning approach: the regex is applied to the entire file contents rather than actual import declarations.

That means an import-shaped path inside a comment or string literal could be counted as a consumer reference and, in `--superseded-enforce` mode, contribute to a false failure. Could we constrain detection to actual Go imports / JS/TS import syntax (or otherwise add tests proving comments and string literals are ignored)?

emphor11 (changes requested):
I reviewed the superseded-construct audit end-to-end, including schema indexing, consumer scanning, generated Go import mappings, and the enforcement path.

The overall direction looks useful, but I found several cases where an incomplete scan or unsupported import mapping can produce an incorrect audit result, including false-clean enforcement. I've left the main issues inline above.

The multi-hop/dangling handling and the general report structure look good; the main concern is making sure a "clean" result is trustworthy before this is used in enforcement mode.

validation/audit.go
Himanshujha7: Hi @emphor11 
Thanks for the detailed review.
You're right, and thanks for catching this.
I went with your second suggestion rather than propagating the error, since
hard-failing would make the report useless in awareness mode over a single bad
file. The index now records what it couldn't read, the report warns about it,
and --superseded-enforce exits 3 for an incomplete audit so it's distinguishable
from an actual finding which could be taken as a good practice. I also made it error out if it indexes zero constructs, a missing schemas/constructs used to return nil and report "0 of 0", which was
the worst version of this in my opinion.

validation/superseded_test.go
Himanshujha7: Yep, you're right.
Go now goes through go/parser in ImportsOnly mode, so only actual import
declarations count and comments and string literals can't contribute at all. For
TS/JS there's no parser in the stdlib, so I've constrained it to the import,
export-from, dynamic import and require forms instead of scanning the file text.
Files that fail to parse get recorded and warned about rather than being treated
as having no imports.

Added tests for both directions too by making comments and string literals ignored, and
the real import forms (grouped, aliased, blank, multi-line, re-export, dynamic,
require) are detected.

validation/source_tree.go
Himanshujha7: I checked before fixing: pointing `--meshery-repo` at a path that
doesn't exist gave a full table of "not used" and exited 0, so a typo'd path
looked identical to a clean consumer.
Right now, I have made a few changes and `WalkFiltered `now returns what it couldn't read instead of swallowing it, and a consumer yielding zero source files counts as unscanned rather than clean.
`Report mode` prints the gaps above the table, and` --superseded-enforce `exits 3
when a requested consumer wasn't fully scanned.

validation/superseded_test.go
Himanshujha7: `meshery/meshery` imports `models/core` in 78 files, and` v1beta1/core` is one of the superseded
constructs. The scanner was missing all of it. I had copied the package-name
mapping from config.js but missed GO_IMPORT_OVERRIDES in
generate-golang.js which is fixed now so scanner now knows about both mappings, so models/core is picked up.  Right now, the report marks it as ambiguous, lists the possible versions, and
points out that one of them is superseded.

Also to mention, I also added a test that reads GO_IMPORT_OVERRIDES straight out of the JS file,
so if that mapping changes again the test fails instead of the audit missing things.

Thanks for the great review I had to check the entire mappings to bring out the faults

Himanshujha7:
Hi @emphor11 
I went through all the reviews, scanned through the mappings to identify the issue pointed out by you and have tried to rectify them over the latest commit.

Thanks for the detailed review.
Please let me know if further changes are needed as well.

cmd/consumer-audit/main.go
```diff
@@ -87,6 +92,70 @@ func main() {
...
+			if !sup.FullyScanned() {
+				fmt.Fprintln(out)
+				for _, repo := range sup.UnscannedRepos() {
+					fmt.Fprintf(out,
+						"superseded-enforce: %s was not fully scanned (%d %s); its result is not conclusive.\n",
+						repo.Repo, len(repo.ScanDefects), pluralize("problem", len(repo.ScanDefects)))
+				}
+				os.Exit(3)
+			}
+			if sup.HasSupersededUsage() {
```
emphor11, line 144: Could we also treat `repo.UnparsedFiles` as an incomplete audit before evaluating `HasSupersededUsage()`?

`scanConsumerForSuperseded` records Go files that fail to parse, but `FullyScanned()` only reflects `ScanDefects`. So an unparsable file can still leave the audit looking clean because its imports were never inspected. In `--superseded-enforce` mode, could we fail with the same exit code `3` for this case before checking `HasSupersededUsage()` / `HasUnresolvedUsage()`?

Himanshujha7: Hi @emphor11 

`FullyScanned() `now covers both, which puts the check ahead of `HasSupersededUsage()` and `HasUnresolvedUsage(`) automatically since enforcement already tests it first `exit 3`, as you suggested. Also fixed the problem count, which would have read `"0 problems"` when an unparsed file was the only fault.
Same case now `exits 3`, and `meshery/meshery `is unaffected: no unparsed files, so
still `exit 2` on real findings.

I tested and ran the whole thing - passes the assertions.
let me know if I would be needed to paste the same over here.

Thanks for such insightful review! loved the feedbacks back n forth.

validation/source_tree.go
```diff
@@ -73,6 +139,80 @@ func (t localTree) Walk(dir string, fn func(path string) error) error {
...
+		// "scanned and found nothing" from "never scanned".
+		if os.IsNotExist(err) {
+			return []ScanDefect{{
+				Path:   filepath.ToSlash(filepath.Join(t.root, dir)),
+				Reason: "directory does not exist; nothing was scanned",
+			}}, nil
+		}
+		return []ScanDefect{{
+			Path:   filepath.ToSlash(filepath.Join(t.root, dir)),
+			Reason: err.Error(),
```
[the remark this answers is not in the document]
Himanshujha7: Had a look, Thanks for the review.
The comment does say "repo-relative where possible", and a missing root is the
case where it isn't possible so it is the same reasoning as the empty-tree defect,
which also identifies the tree rather than a path inside it. Everything below
the root is repo-relative.
I've made that split explicit in the doc comment rather than leaving it to the
hedge, since that's what made this look like a violation.

cmd/consumer-audit/main.go
```diff
@@ -412,6 +482,185 @@ func printTSFindings(out io.Writer, findings []validation.TSFinding) {
...
+		// should be read: "not used" from an unscanned tree means "not
+		// seen".
+		if !repo.FullyScanned() {
+			n := len(repo.ScanDefects) + len(repo.UnparsedFiles)
+			fmt.Fprintf(out, "    WARNING: this consumer was not fully scanned (%d %s).\n",
+				n, pluralize("problem", n))
+			fmt.Fprintln(out, "    Rows below say what was seen, not what exists:")
+			for _, d := range repo.ScanDefects {
+				fmt.Fprintf(out, "      %s\n        %s\n", d.Path, d.Reason)
+			}
```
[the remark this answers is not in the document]
Himanshujha7: Hey, Thanks for the review.
I have tried to address this over the latest PR - the code block lead to duplication of list since I already had an implementation at another segment but currently, There is only one implementation for fullyscanned() and no duplicates are present.

validation/source_tree.go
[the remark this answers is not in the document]
Himanshujha7: Hi,

I have tried to address this in the latest commit - please see through it if helps.

Himanshujha7:
Hi @emphor11 let me know if it 
looks GTG to you now?

validation/source_tree.go
```diff
@@ -73,6 +139,80 @@ func (t localTree) Walk(dir string, fn func(path string) error) error {
 	return nil
 }
 
+func (t localTree) WalkFiltered(dir string, exts []string, fn func(path string) error) ([]ScanDefect, error) {
+	var defects []ScanDefect
+	defectPath := filepath.ToSlash(filepath.Clean(dir))
+	abs := filepath.Join(t.root, filepath.FromSlash(dir))
+	info, err := os.Stat(abs)
```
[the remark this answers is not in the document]
Himanshujha7: Two reasons I'd rather defer this one.

It's pre-existing behaviour, not something this PR introduces. The existing
Walk method on master has the identical os.Stat/WalkDir pattern; I have tried to inehrit the same pattern for WalkFiltered. Fixing only the new method would leave two methods in the same
file disagreeing about symlinked roots, which seems worse than either state.

Also to mention as per what I could check is no consumer is symlinked today and CI does real checkouts so I'd rather fix both together in a follow-up than half of it here. Happy to open that issue.

Also, I could be wrong about this - please let me know

CC @emphor11

Himanshujha7:
Hi @emphor11 @marblom007 

Could you please check and review this PR and let me know if it needs any corrections from my end?
Kindly let me know, since it has been pending for a while now.

emphor11:
@Himanshujha7 Could we prevent an ambiguous import from also producing a NotUsed row for the same family?

recordAmbiguous() returns before record(), so a version-erased path such as models/core never populates files[family]. The later !found branch therefore adds NotUsed even though the consumer did reference that family; only the exact version is unknown.

Could we track families that produced an ambiguous resolution and skip their NotUsed row? It would also be useful to add an assertion to the models/core test so this contradictory state stays covered.

PR #1152, opened 2026-08-10
fix: add avatarUrl to v1beta3 EventResult

### LINKED ISSUE #1145: v1beta3 EventResult omits `avatarUrl`, which the server emits alongside the other joined user fields
## Summary

The canonical v1beta3 `EventResult` (and the inline copy inside `EventsPage.data[]`) models the joined user fields `firstName`, `lastName`, `email` and `provider`, but not `avatarUrl` - which meshery-cloud does emit on the same object.

`meshery-cloud/server/models/event_tracker.go`:

```go
	Provider  string `json:"provider" db:"provider"`
	Email     string `json:"email" db:"email"`
	FirstName string `json:"firstName" db:"first_name"`
	LastName  string `json:"lastName" db:"last_name"`
	AvatarUrl string `json:"avatarUrl" db:"avatar_url"`
```

So a consumer generating its types from `constructs/v1beta3/event/Event.yml` cannot name the field it needs to render an activity row's avatar, and has to bolt a local extension onto the canonical type.

## Where it bites

Sistent's workspace recent-activity modal and the workspace activity dashboard widget both render the actor's avatar next to each event. layer5io/sistent#1780 repoints those rows at `EventResult` and carries `avatarUrl` as an explicitly documented local addition pointing here; it should be deleted once the canonical models the field.

The other four joined user fields are already present in the canonical, so this reads as an omission rather than a deliberate exclusion - `avatarUrl` is not personally identifying in a way that `email`, which *is* modelled, is not.

## Suggested resolution

Add `avatarUrl` (`type: string`, `format: uri`, optional) to `EventResult` in `schemas/constructs/v1beta3/event/`, alongside the existing joined user fields, and regenerate.

suii2210:
## Notes for Reviewers

This PR fixes #1145.

### What changed

- Adds the missing optional `avatarUrl` field to the v1beta3 `EventResult` schema.
- Defines `avatarUrl` as a `string` with `format: uri`.
- Regenerates the affected Go and TypeScript models.
- Regenerates the RTK Cloud client.

### Why

Meshery Cloud emits `avatarUrl` alongside the existing joined user fields (`firstName`, `lastName`, `email`, and `provider`), but the canonical v1beta3 `EventResult` schema did not expose this field.

This caused schema consumers to maintain a local extension when rendering user avatars.

### Validation

- `make build` 
- Schema validation 
- RTK generation tests 
- `git diff --check` 

---

### DIFF
--- a/models/v1beta3/event/event.go
+++ b/models/v1beta3/event/event.go
@@ -66,6 +66,9 @@ type EventResult struct {
 	// Action The action of the event.
 	Action string `json:"action" yaml:"action"`
 
+	// AvatarUrl URL to the avatar image of the user associated with the event.
+	AvatarUrl *string `json:"avatarUrl,omitempty" yaml:"avatarUrl,omitempty"`
+
 	// Category The category of the event.
 	Category string `json:"category" yaml:"category"`
 
--- a/schemas/constructs/v1beta3/event/event.yaml
+++ b/schemas/constructs/v1beta3/event/event.yaml
@@ -42,6 +42,11 @@ properties:
     type: string
     description: The last name of the user associated with the event.
     maxLength: 500
+  avatarUrl:
+    type: string
+    format: uri
+    maxLength: 500
+    description: URL to the avatar image of the user associated with the event.
   email:
     $ref: ../../v1beta2/core/api.yml#/components/schemas/Email
     description: Email address of the user associated with the event.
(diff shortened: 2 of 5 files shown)


pontusringblom (commented):
Thanks! Is v1beta3 the current version of events? If so, great. If not, then lets use v1beta2 (or whatever the latest is), since we don't consider this change worthy of a version bump.

suii2210:
> Thanks! Is v1beta3 the current version of events? If so, great. If not, then lets use v1beta2 (or whatever the latest is), since we don't consider this change worthy of a version bump.

v1beta3 already exists for the Event schema on master (schemas/constructs/v1beta3/event/api.yml and event.yaml), and the Event migration work also targets v1beta3. So this PR is updating the existing Event version rather than introducing a new version bump.

PR #1163, opened 2026-08-16
test(event): add contract-pinning validation tests for Event construct

NSTKrishna:
## Description

Adds contract-pinning validation tests for the Event construct (Notification Center) and fixes a field-name drift in the event template. Modeled on the Subscription migration (`validation/subscription_operations_test.go`).

### Changes

**`validation/event_operations_test.go`** (new)

Pins the contract of all 10 event operations with 8 test functions (38 subtests):

| Test | What it pins |
|------|-------------|
| `TestEventOperationsAreDeclared` | Every `operationId` resolves at its declared path+method |
| `TestEventPathsAreDeclaredUnderTheApiPrefix` | All paths carry `/api` prefix — missing prefix generates a client that 404s at runtime |
| `TestEventDualClientOperationsAreAnnotatedCorrectly` | `x-internal` is `[cloud meshery]` on the 5 dual-client mutations, `[cloud]` on the 5 cloud-only queries |
| `TestEventMutationsUseCorrectRequestBody` | Each mutation references its dedicated schema with `required: true` |
| `TestEventMutationsDoNotReferenceEntitySchema` | No mutation references `EventResult` (entity schema) |
| `TestEventPayloadProperties` | `EventPayload` has exactly `{owner, category, action, description}` — no server-generated fields |
| `TestEventEntitySchemaIsResponseOnly` | `event.yaml` has `additionalProperties: false` |
| `TestEventResponseCodes` | Each operation's declared status codes match expectations |

**`schemas/constructs/v1beta3/event/templates/event_template.json`** (modified)

- Renamed `userId` → `owner` to match the entity schema property name
- The YAML template already used `owner`; the JSON template was never updated
- With `additionalProperties: false` on the entity schema, `userId` is an invalid extra property and the required `owner` was absent

### RTK Coverage Audit

As part of this work, all 10 event operations were audited against `typescript/rtk/cloud.ts` and `typescript/rtk/meshery.ts`. **No gaps or client drift were found** — all operations are present in the correct clients with matching URLs, methods, and parameter names.

### Testing

```
go test ./validation/... -run TestEvent -v   # 38/38 subtests pass
make validate-schemas                         # no blocking violations
go test ./...                                 # all packages pass
npm run build                                 # build success
```

### No consumer PRs needed

No generated artifacts were modified (`models/`, `typescript/generated/`, `typescript/rtk/`). This PR adds tests and fixes a template only.

### Deferred follow-ups (out of scope for this PR)

- Adding `id`, `status`, `updatedAt`, `deletedAt` to `event.yaml` (changes generated types, requires downstream coordination)
- Adding explicit `security: [jwt: []]` to the 9 operations that omit it
- Authorization middleware table (requires inspecting meshery-cloud `router.go` for which event routes carry `AuthorizationMiddlewareForOrgAdminAndOrgOwner`)

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated event templates to use the correct owner field, with a zero UUID default.

* **Tests**
  * Added comprehensive contract coverage for all event operations, including request validation, response schemas, endpoint paths, operation identifiers, internal annotations, and status codes.


vedant21-ctr (changes requested):
Two small things:

Header comment says "eight paths" — it's actually 9 distinct paths in eventOperations (only eventsRootPath is shared).

Templates include systemId, but EventPayload doesn't pin it. Assuming that's because systemId is server-set like the entity ID — can you confirm in a reply so it's on record?

Everything else looks good to me.

NSTKrishna:
> Two small things:
> 
> Header comment says "eight paths" — it's actually 9 distinct paths in eventOperations (only eventsRootPath is shared).
> 
> Templates include systemId, but EventPayload doesn't pin it. Assuming that's because systemId is server-set like the entity ID — can you confirm in a reply so it's on record?
> 
> Everything else looks good to me.

yes, you're exactly right about systemId and it's a server-set field identifying the event source, just like the entity ID

hiyach28 (approved):
Reviewed this carefully against the Event schema and existing validation patterns. I also ran the schema validation, consumer audit, Event tests, and code generation checks; everything passed with no unexpected downstream changes.

LGTM!

Maanvi212006 (approved):
LGTM! Reviewed the branch end to end — ran the event tests, full suite, schema validation, and build; all green, and confirmed no generated clients were touched. Template fix is correct (real `userId`/`owner` drift), and the new tests lock down the contract well.

PR #1169, opened 2026-08-18
fix(registration): adopt existing model ID; make relationship Create idempotent

### LINKED ISSUE #1168: Registering relationships into an existing model orphans them (nil model_id); relationship registration is non-idempotent
### Current Behavior

Two defects in the registration helpers compound into silent, systemic registry corruption in every Meshery deployment:

**1. `ModelDefinition.Create` (v1beta1 `model_helper.go`) loses the model ID on the already-exists path.**

```go
err = db.First(&model, "id = ? and connection_id = ?", modelID, hostID).Error
...
if err == gorm.ErrRecordNotFound {
    m.ID = modelID          // assigned only on the insert path
    ...
    return m.ID, nil
}
return model.ID, nil        // found: returned, but m.ID is NEVER set
```

meshkit's `registration.register()` subsequently does `rel.ModelId = &model.ID` for every relationship in the package. When the model already existed (every seeding run after the first; every `mesheryctl model import` into a model the server already knows), `model.ID` is still the nil UUID from the JSON, so **every relationship registers with `model_id = 00000000-...`** - orphaned, invisible to every model-scoped query (`/api/meshmodels/models/{name}/relationships`, `mesheryctl relationship view <model>`, the evaluation engine's model-filtered relationship loading).

**2. `RelationshipDefinition.GenerateID` (v1alpha3 `relationship_helper.go`) is `uuid.NewV4()`.**

Models are content-addressed and idempotent; relationships are not. Every registration inserts a brand-new row - no existing-row check, no dedupe - so every server restart re-inserts the full relationship corpus.

### Evidence (local dev server, seeded from meshery/meshery master)

```
sqlite> SELECT count(*) FROM relationship_definition_dbs;
11256
sqlite> SELECT count(*) FROM relationship_definition_dbs
        WHERE model_id='00000000-0000-0000-0000-000000000000';
10262   -- 91% of all registered relationships are orphaned
```

Reproduction of the import flow: `mesheryctl model build` + `model import` of a package carrying 3 relationship definitions for the (already-registered) cert-manager model reports `SUMMARY: Imported model cert-manager (3 relationships)` with `errRelCount: 0` - and `/api/meshmodels/models/cert-manager/relationships` still returns `total_count: 0`. The rows exist, orphaned. The documented "bundle your private relationships in a custom model and import it" flow is broken for any model the server has already seen, and the API response claims success.

### Desired Behavior

1. The already-exists branch assigns the found identity back to the receiver (`m.ID = model.ID`, plus `CategoryId`/`RegistrantId` for struct consistency) so callers holding the struct see the real ID.
2. Relationship registration becomes idempotent the same way model registration is: content-address the definition (hash of schemaVersion/kind/type/subType/version/model reference/selectors, excluding volatile fields) and return the existing row's ID instead of inserting a duplicate.

Existing deployments retain the orphaned/duplicate rows already accumulated; a cleanup migration in meshery/meshery is follow-up work (tracked separately).

### Environment
- meshery/schemas master, meshkit master, Meshery Server v1.0.66 (local provider)

leecalcote:
**Description**

This PR fixes meshery/schemas#1168.

Two compounding defects in the registration helpers silently corrupt the relationship registry in every Meshery deployment:

1. **`ModelDefinition.Create` (v1beta1) loses the model identity on the already-exists path.** It returned the found row's ID without assigning it to the receiver. meshkit's `registration.register()` reads `m.ID` after `Create` to stamp `ModelId` onto every component and relationship in the package, so whenever the model was already registered - every seeding run after the first, every `mesheryctl model import` into a known model - **all relationships registered with `model_id = nil UUID`**, orphaned from every model-scoped query and from evaluation. On a local dev server, 10,262 of 11,256 relationship rows (91%) are orphaned this way.

2. **`RelationshipDefinition.GenerateID` (v1alpha3) was `uuid.NewV4()`** - no idempotency, so every restart re-inserted the entire relationship corpus as fresh rows.

**Changes**

- Exists path of `ModelDefinition.Create` adopts the persisted identity onto the receiver (`ID`, `CategoryId`, `RegistrantId`).
- `RelationshipDefinition.GenerateID` content-addresses the definition from its semantic coordinates (schemaVersion, version, kind, type, subType, modelId, evaluationQuery, selectors - volatile/cosmetic fields excluded), mirroring `ModelDefinition.GenerateID`.
- `RelationshipDefinition.Create` returns the existing row's ID instead of inserting a duplicate, serialized by a creation lock the way model creation is.
- Regression tests: receiver-ID adoption on the exists path, content-addressed hashing, idempotent Create (in-memory sqlite via meshkit's database handler).

**Reproduction that motivated this** (full detail in #1168): `mesheryctl model build` + `model import` of a package carrying 3 cert-manager relationship definitions reports success with `errRelCount: 0`, and `/api/meshmodels/models/cert-manager/relationships` still returns `total_count: 0`.

**Out of scope**

Rows already orphaned/duplicated in existing deployments; that cleanup is a meshery/meshery migration (follow-up).

**Verification**

- `go build ./...`, `go test ./...` - clean
- `go vet` on the touched packages - clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Relationship registrations now generate consistent IDs and prevent duplicate records when the same definition is submitted multiple times, including concurrent registrations.
  * Existing model records now correctly return their persisted identifiers and associated metadata during creation.
  * Duplicate model registrations no longer fail during concurrent creation.

* **Documentation**
  * Added guidance on registration identity, duplicate handling, persisted identity adoption, and compatibility with existing records.

* **Tests**
  * Added coverage for deterministic IDs, idempotent registration across database connections, conflict handling, and model creation with existing records.

### DIFF
--- a/models/v1alpha3/relationship/relationship_helper.go
+++ b/models/v1alpha3/relationship/relationship_helper.go
@@ -13,6 +13,7 @@ import (
 	"github.com/meshery/meshkit/database"
 	"github.com/meshery/meshkit/models/meshmodel/entity"
 	"github.com/meshery/meshkit/utils"
+	"github.com/meshery/schemas/models/core"
 	"gorm.io/gorm"
 	"gorm.io/gorm/clause"
 )
@@ -48,15 +49,33 @@ func (r RelationshipDefinition) Type() entity.EntityType {
 	return entity.RelationshipDefinition
 }
 
+// relationshipIdentity is the canonical set of fields hashed by GenerateID.
+// It is a dedicated struct with its own JSON tags rather than a
+// RelationshipDefinition so the identity cannot silently drift with the
+// entity's wire representation: ModelId is json:"-" on the entity, which
+// would drop it from a Marshal of the entity and let equivalent relationships
+// from different models collide on one ID. Every field here is always
+// serialized, so field order and presence are fixed by this struct alone.
+type relationshipIdentity struct {
+	SchemaVersion    core.VersionString         `json:"schemaVersion"`
+	Version          core.SemverString          `json:"version"`
+	Kind             RelationshipDefinitionKind `json:"kind"`
+	RelationshipType string                     `json:"type"`
+	SubType          string                     `json:"subType"`
+	ModelId          *core.Uuid                 `json:"modelId"`
+	EvaluationQuery  *string                    `json:"evaluationQuery"`
+	Selectors        *SelectorSet               `json:"selectors"`
+}
+
 // GenerateID content-addresses the definition from its semantic coordinates,
 // mirroring ModelDefinition.GenerateID. Registering the same definition twice
 // (server re-seeding on restart, re-importing a model package) therefore
 // resolves to the same ID instead of minting a fresh random UUID per
 // registration, which is what let duplicate rows accumulate unboundedly.
-// Volatile and cosmetic fields (ID, Status, Metadata) are deliberately
-// excluded, matching the model identifier's philosophy.
+// Volatile and cosmetic fields (ID, Status, Metadata, Capabilities) are
+// deliberately excluded, matching the model identifier's philosophy.
 func (r *RelationshipDefinition) GenerateID() (uuid.UUID, error) {
-	relationshipIdentifier := RelationshipDefinition{
+	byt, err := json.Marshal(relationshipIdentity{
 		SchemaVersion:    r.SchemaVersion,
 		Version:          r.Version,
 		Kind:             r.Kind,
@@ -65,8 +84,7 @@ func (r *RelationshipDefinition) GenerateID() (uuid.UUID, error) {
 		ModelId:          r.ModelId,
 		EvaluationQuery:  r.EvaluationQuery,
 		Selectors:        r.Selectors,
-	}
-	byt, err := json.Marshal(relationshipIdentifier)
+	})
 	if err != nil {
 		return uuid.UUID{}, err
 	}
--- a/models/v1alpha3/relationship/relationship_helper.go
+++ b/models/v1alpha3/relationship/relationship_helper.go
@@ -110,11 +110,18 @@ func (r *RelationshipDefinition) Create(db *database.Handler, hostID uuid.UUID) 
 	// ON CONFLICT DO NOTHING turns the loser's duplicate-key error into a
 	// no-op, and because the ID is content-addressed the existing row is the
 	// same definition, so returning the computed ID is correct either way.
-	err = db.Omit(clause.Associations).Clauses(clause.OnConflict{DoNothing: true}).Create(&r).Error
-	if err != nil {
+	if err := r.insertIgnoringConflict(db); err != nil {
 		return uuid.UUID{}, err
 	}
 	return id, nil
+}
+
+// insertIgnoringConflict persists the definition, treating a primary-key
+// collision as a successful no-op. It is the single seam Create relies on to
+// survive the cross-process race described above, and is unexported so tests
+// can drive the conflict path directly without racing two connections.
+func (r *RelationshipDefinition) insertIgnoringConflict(db *database.Handler) error {
+	return db.Omit(clause.Associations).Clauses(clause.OnConflict{DoNothing: true}).Create(r).Error
 }
 
 func (r *RelationshipDefinition) UpdateStatus(db *database.Handler, status entity.EntityStatus) error {
--- a/models/v1alpha3/relationship/relationship_helper.go
+++ b/models/v1alpha3/relationship/relationship_helper.go
@@ -104,7 +104,13 @@ func (r *RelationshipDefinition) Create(db *database.Handler, hostID uuid.UUID) 
 		return existing.ID, nil
 	}
 
-	err = db.Omit(clause.Associations).Create(&r).Error
+	// The mutex above only serializes callers within one process. Concurrent
+	// registrations from separate processes sharing a database can both pass
+	// the exists check, so the insert itself must tolerate losing that race:
+	// ON CONFLICT DO NOTHING turns the loser's duplicate-key error into a
+	// no-op, and because the ID is content-addressed the existing row is the
+	// same definition, so returning the computed ID is correct either way.
+	err = db.Omit(clause.Associations).Clauses(clause.OnConflict{DoNothing: true}).Create(&r).Error
 	if err != nil {
 		return uuid.UUID{}, err
 	}
--- a/models/v1alpha3/relationship/relationship_helper.go
+++ b/models/v1alpha3/relationship/relationship_helper.go
@@ -2,16 +2,25 @@
 package relationship
 
 import (
+	"crypto/md5"
+	"encoding/json"
 	"fmt"
 	"path/filepath"
 	"strings"
+	"sync"
 
 	"github.com/gofrs/uuid"
 	"github.com/meshery/meshkit/database"
 	"github.com/meshery/meshkit/models/meshmodel/entity"
 	"github.com/meshery/meshkit/utils"
+	"gorm.io/gorm"
 	"gorm.io/gorm/clause"
 )
+
+// relationshipCreationLock serializes the exists-check-then-insert in Create so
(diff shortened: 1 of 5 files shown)


leecalcote:
@coderabbitai review

(The earlier automatic review was skipped by the hourly review limit - requesting it explicitly so this PR gets a bot pass. The consumer audit report above is informational and shows no endpoint drift from this change.)

models/v1alpha3/relationship/relationship_helper.go
```diff
@@ -39,8 +48,31 @@ func (r RelationshipDefinition) Type() entity.EntityType {
 	return entity.RelationshipDefinition
 }
 
+// GenerateID content-addresses the definition from its semantic coordinates,
+// mirroring ModelDefinition.GenerateID. Registering the same definition twice
+// (server re-seeding on restart, re-importing a model package) therefore
+// resolves to the same ID instead of minting a fresh random UUID per
+// registration, which is what let duplicate rows accumulate unboundedly.
+// Volatile and cosmetic fields (ID, Status, Metadata) are deliberately
+// excluded, matching the model identifier's philosophy.
```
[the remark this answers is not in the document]
leecalcote: Documented in 63ec1aa2: docs/registration-identity.md defines the identity fields hashed for models and relationships, the no-op semantics of re-registration, receiver adoption, cross-process behavior, and the compatibility posture toward rows created before content-addressing.

models/v1alpha3/relationship/relationship_helper.go
```diff
@@ -58,11 +90,25 @@ func (r *RelationshipDefinition) Create(db *database.Handler, hostID uuid.UUID)
...
+	}
+	if err == nil {
+		// The definition is already registered; re-registration is a no-op,
+		// the same way ModelDefinition.Create treats an existing model.
+		return existing.ID, nil
+	}
+
 	err = db.Omit(clause.Associations).Create(&r).Error
 	if err != nil {
 		return uuid.UUID{}, err
```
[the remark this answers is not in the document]
leecalcote: Fixed in 63ec1aa2: the insert now carries ON CONFLICT DO NOTHING, so a registration that loses the cross-process race becomes a no-op instead of a duplicate-key error - and because the ID is content-addressed, the winner's row is the same definition, making the computed ID correct in both branches without a re-load. Covered by a two-connection idempotency test against one sqlite file (an in-process goroutine pair cannot exercise the cross-process window since the package mutex serializes them; the test validates the contract end to end across independent connections).

models/v1beta1/model/model_helper.go
```diff
@@ -103,6 +103,14 @@ func (m *ModelDefinition) Create(db *database.Handler, hostID uuid.UUID) (uuid.U
...
 		return m.ID, nil
 	}
+	// The model already exists: adopt the persisted identity onto the receiver.
+	// Callers (meshkit registration.register) read m.ID after Create to stamp
+	// ModelId onto every component and relationship in the package; leaving the
+	// receiver's ID untouched here made all of them register with the nil UUID,
+	// orphaning them from every model-scoped query.
+	m.ID = model.ID
+	m.CategoryId = model.CategoryId
+	m.RegistrantId = hostID
```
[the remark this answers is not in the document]
leecalcote: Added in 63ec1aa2: the exists-path test now asserts CategoryId matches the first registration's persisted value and RegistrantId matches hostID, alongside the ID assertion. Note the test file is hand-authored, not generated - model_helper_test.go was introduced by this PR - so the assertions live there directly.

leecalcote:
All three review findings are addressed in 63ec1aa2 (identity-contract docs, ON CONFLICT DO NOTHING with a two-connection test, receiver-field assertions) with individual replies on each thread. The walkthrough and consumer-audit report carry no further asks.

models/v1alpha3/relationship/relationship_helper_test.go
```diff
@@ -0,0 +1,142 @@
...
+
+	hostID := uuid.Must(uuid.NewV4())
+	defA := testDefinition("reference")
+	defB := defA
+
+	firstID, err := defA.Create(&first, hostID)
+	if err != nil {
+		t.Fatalf("create on first connection: %v", err)
+	}
+	secondID, err := defB.Create(&second, hostID)
```
ishwar170695, line 155: One question about this test: since these calls run sequentially, wouldn't the second one find the row during `db.First` and return before reaching `ON CONFLICT DO NOTHING`? Wondering if there's a way to exercise that conflict path as well.

models/v1beta1/model/model_helper.go
ishwar170695: Small question here: since we're adopting the persisted identity onto the receiver, should this be `model.RegistrantId` instead of `hostID`? The test also currently checks against `hostID`, so using a different host for the second registration might help verify that we're actually adopting the stored value.

leecalcote: @ishwar170695, this is a well-placed question.

models/v1alpha3/relationship/relationship_helper_test.go
leecalcote: Good catch - sequential calls never reach it. Extracted the ON CONFLICT DO NOTHING insert into insertIgnoringConflict and added TestInsertIgnoringConflictIsNoOpOnDuplicate, which inserts a row, proves a plain insert of the same ID collides, then drives the conflict-tolerant insert and asserts no error and exactly one row.

models/v1beta1/model/model_helper.go
leecalcote: Agreed - switched to model.RegistrantId so every adopted field mirrors the persisted row. A different host on the second registration can't verify it, though: the lookup is scoped by connection_id = hostID, so a different host is a miss rather than a match. The test now loads the persisted row and asserts the receiver's RegistrantId equals the stored value rather than the argument.

models/v1alpha3/relationship/relationship_helper.go
```diff
@@ -1,84 +1,143 @@
...
+	relationshipIdentifier := RelationshipDefinition{
+		SchemaVersion:    r.SchemaVersion,
+		Version:          r.Version,
+		Kind:             r.Kind,
+		RelationshipType: r.RelationshipType,
+		SubType:          r.SubType,
+		ModelId:          r.ModelId,
+		EvaluationQuery:  r.EvaluationQuery,
+		Selectors:        r.Selectors,
+	}
```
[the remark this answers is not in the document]
leecalcote: ModelId: confirmed and fixed - the entity tags it json:"-", so it was dropped from the hash. GenerateID now marshals a dedicated relationshipIdentity struct with explicit tags, and TestGenerateIDIsContentAddressed covers the cross-model case. Capabilities: declined - they are behavioral metadata in the same class as metadata/styles, not identity; a definition differing only in capabilities is the same relationship (test asserts this, doc lists it as excluded). Model hostID: declined for this PR - ModelDefinition.GenerateID predates this change, registrant is already part of the hash, and adding hostID would re-identify every registered model; noted in docs/registration-identity.md as a deliberate host-agnostic design.

models/v1beta1/model/model_helper.go
```diff
@@ -1,175 +1,187 @@
...
+		return modelID, err
+	}
+
+	var model ModelDefinition
+	if m.Name == "" {
+		return uuid.UUID{}, fmt.Errorf("empty or invalid model name passed")
+	}
+	modelCreationLock.Lock()
+	defer modelCreationLock.Unlock()
+	err = db.First(&model, "id = ? and connection_id = ?", modelID, hostID).Error
```
ishwar170695, line 83: One question here: `GenerateID()` is host-agnostic, and the new `registration-identity.md` now explicitly documents model identity as host-agnostic, but this lookup is still scoped by `connection_id`.

Could the same model registered from another host therefore miss the existing row here and then hit a primary-key conflict when attempting to insert the same deterministic `modelID`?

Is cross-host registration intentionally unsupported, or should this lookup be global by `id` to match the documented identity behavior?

PR #1170, opened 2026-08-20
fix(academy): change registrationCount to int64 and remove omitempty

### LINKED ISSUE #1116: v1beta3 academy: registrationCount is `required` yet `omitempty`, and typed `number` (float32) for a count
## Summary

`registrationCount` is declared **`required`** in two v1beta3 academy schemas, yet its generated tags carry **`omitempty`**. Those two statements contradict each other: the schema says the field is always present, while the Go marshaller drops it whenever the value is `0`. It is also typed `number`, which generates `float32` for what is a count.

This is a latent wire-contract hazard. **It did not cause any known outage** - see the context section below, which is explicit about what did.

## Where

`schemas/constructs/v1beta3/academy/api.yml`, in **both** of these schemas:

- `SingleAcademyCurriculaResponse` (~L857-871)
- `AcademyCurriculaWithMetrics` (~L941-954)

Both read:

```yaml
required:
- registrationCount
properties:
  registrationCount:
    type: number
    description: Number of registrations associated with this curriculum.
    minimum: 0
    x-oapi-codegen-extra-tags:
      db: registration_count,omitempty
      json: registrationCount,omitempty
```

Which generates (`models/v1beta3/academy/academy.go`):

```go
RegistrationCount float32 `db:"registration_count,omitempty" json:"registrationCount,omitempty" yaml:"registrationCount,omitempty"`
```

## Three distinct problems

1. **`required` + `omitempty` is self-contradictory.** A curricula with zero registrations serializes a payload that violates the schema's own `required` constraint, because the field is omitted entirely. Any consumer validating against this schema can reject a response the server considers valid.

2. **A genuine zero is indistinguishable from an absent field.** With `omitempty`, "this content has 0 registrations" and "this field was never populated" produce byte-identical payloads. That is a real diagnostic hazard for a metric: it makes a broken count and an honestly-empty count look the same on the wire, so a consumer cannot tell a working pipeline reporting zero from a silently failing one.

3. **`type: number` is the wrong type for a count.** It generates `float32`. A registration count is a non-negative whole number and cannot be fractional; `float32` additionally loses exactness above 2^24. `type: integer` is correct on its own terms, independent of the `omitempty` question.

## Suggested fix

```yaml
registrationCount:
  type: integer
  format: int64
  description: Number of registrations associated with this curriculum.
  minimum: 0
  x-oapi-codegen-extra-tags:
    db: registration_count
    json: registrationCount
```

Dropping `omitempty` makes the emitted payload match the `required` declaration and lets a real `0` serialize. Apply to both schemas.

## Consumer impact

Consumers currently reading this field must keep a `|| 0` / `?? 0` fallback until the change ships, because today an absent key legitimately means zero. After the change the key is always present, so the fallback becomes redundant but stays harmless. Known consumer: layer5io/meshery-cloud, `ui/components/academy/InstructorConsole/overview.tsx`.

## Context - how this was found, and what it was NOT

Found while fixing a Meshery Cloud bug where the Instructors Console overview reported "0 registrations" for content whose learners tab listed 230.

**The root cause of that bug was a separate issue: a wire-key casing rename.** The academy migration to v1beta3 renamed the JSON key from `registration_count` (v1beta1, snake_case) to `registrationCount` (camelCase). The Cloud UI widget had been written against v1beta1 and kept reading the snake_case name, so it got `undefined` and its `|| 0` fallback rendered a plausible "0 registrations". That has been fixed on the consumer side by reading the correct camelCase key; the schema was correct there and needed no change.

The `omitempty` / `required` / `float32` problems described above are a **separate latent hazard noticed while diagnosing that bug**. They did not cause it. They did, however, make it harder to diagnose: because a real `0` is omitted from the payload, "the count was computed as zero" and "the count never arrived" were indistinguishable on the wire, which is a defect worth removing on its own merits.

dhruveshmishra:
#### Description
Fixes #1116
This PR resolves a schema definition contradiction and data type issue for `registrationCount` in the v1beta3 Academy OpenAPI schemas (`SingleAcademyCurriculaResponse` and `AcademyCurriculaWithMetrics`):
1. **Removed `omitempty` tag**: The field is marked as `required`, but generated `omitempty` struct tags caused Go's JSON encoder to drop the field whenever `registrationCount` was `0`. Removing `omitempty` ensures that a count of `0` is always serialized on the wire as `"registrationCount": 0`.
2. **Correct Data Type**: Changed `type: number` (which generated `float32`) to `type: integer` with `format: int64`, correctly representing count values as whole integers.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Registration counts are now represented as whole numbers for greater accuracy and consistency.
  * Registration count fields are now always included in academy curriculum responses, including when the count is zero.
  * Updated API definitions to reflect the revised registration count format.


emphor11 (approved):
Ty for this thorough fix

ishwar170695 (commented):
Small question: since `models/` and `typescript/generated/` are generated and committed by automation on `master`, should we leave these generated artifacts out of the PR and let the post-merge workflow handle them?

dhruveshmishra:
@ishwar170695 Yeah , good catch, you're right! I've removed the generated files and updated the PR to only include the schema changes.

ishwar170695 (approved):
LGTM

PR #1181, opened 2026-09-02
[Models] fix: update ModelDefinition receiver ID on existing-model path

AnimeshPriyanshu:
## Description

Fix `ModelDefinition.Create()` when re-registering an existing model.

When `Create()` finds an already-registered model, it returns the existing
`model.ID` but previously did not update the receiver's `m.ID`. Callers that
subsequently use `m.ID` as a foreign key could therefore write the zero UUID
instead of the real database model ID.

This can cause relationships registered during model re-registration to
become orphaned and invisible to registry queries.

### Fix

Update `m.ID` with the existing model's ID before returning.

### Testing

Added a regression test covering the existing-model path and verifying that
the receiver's `ID` is updated to the existing database model ID.

Tests:
- `go test -count=1 -v -run TestCreate_ExistingModelUpdatesReceiverID ./models/v1beta1/model/`
- `go test ./models/v1beta1/model/...`

### Context

This issue was discovered while testing:
https://github.com/meshery/meshery/pull/21472

The Meshery PR itself remains separate; this PR fixes the underlying
`meshery/schemas` model re-registration issue.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed model imports so the model ID is correctly retained when an existing model is detected.
  * Prevented inconsistent model identification during repeated imports.

### DIFF
--- a/models/v1beta1/model/model_helper.go
+++ b/models/v1beta1/model/model_helper.go
@@ -103,6 +103,7 @@ func (m *ModelDefinition) Create(db *database.Handler, hostID uuid.UUID) (uuid.U
 		}
 		return m.ID, nil
 	}
+	m.ID = model.ID
 	return model.ID, nil
 }
 
--- a/models/v1beta1/model/model_helper_test.go
+++ b/models/v1beta1/model/model_helper_test.go
@@ -1,0 +1,129 @@
+package model
+
+import (
+	"testing"
+
+	"github.com/gofrs/uuid"
+	"github.com/meshery/meshkit/database"
+	category "github.com/meshery/schemas/models/v1beta1/category"
+	connection "github.com/meshery/schemas/models/v1beta1/connection"
+	"gorm.io/driver/sqlite"
+	"gorm.io/gorm"
+)
+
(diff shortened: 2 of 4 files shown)


ishwar170695 (commented):
Should we also update `CategoryId` and `RegistrantId` along with `m.ID` here? The existing DB record has those values, but the receiver still keeps zero values on the existing-model path.

Also, `v1beta2` has the same `Create()` path without this receiver update. Should we apply the same fix there for version parity?

AnimeshPriyanshu:
@ishwar170695 
Yeah, I checked both points.

For CategoryId and RegistrantId, I’d leave them as-is for now. I couldn’t find any current caller that relies on those receiver fields after Create() returns. The concrete issue we found was with m.ID, where a zero/stale ID can cause relationships to be associated incorrectly. So I’d keep the fix focused rather than changing fields without a demonstrated need.

For v1beta2, yes, I applied the same m.ID = model.ID fix there for version parity and added a regression test. Both the focused test and the full v1beta2/model tests are passing.

PR #1183, opened 2026-09-03
Give updateView its own partial-update payload

### LINKED ISSUE #1147: updateView is a partial-update PUT but reuses the create payload, so a visibility-only update has no schema-valid body
## Problem

`updateView` is a partial-update `PUT`, but it reuses the create request body, so the schema mandates a field the operation has no business requiring. A caller changing only a view's visibility has no valid way to express that request.

`createView` and `updateView` both reference the same request body:

```yaml
# schemas/constructs/v1beta2/view/api.yml
  requestBodies:
    viewPayload:
      description: Body for creating or updating a view
      ...
            $ref: "#/components/schemas/ViewPayload"

paths:
  /api/content/views:
    post:
      operationId: createView
      requestBody:
        $ref: "#/components/requestBodies/viewPayload"     # line 343
  /api/content/views/{viewId}:
    put:
      operationId: updateView
      requestBody:
        $ref: "#/components/requestBodies/viewPayload"     # line 466
```

and that payload makes `name` both required and non-empty:

```yaml
    ViewPayload:
      type: object
      description: Payload for creating or updating a view.
      required:
        - name
      properties:
        name:
          type: string
          description: Display name of the view.
          maxLength: 255
          minLength: 1
```

Same in `v1beta1`.

## The provider is explicitly a partial update

`meshery-cloud`'s `UpdateView` retains the stored value for **every** field the caller omits - not just `name`:

```go
// meshery-cloud server/handlers/meshery_views.go
if viewPayload.Name == "" {
	viewPayload.Name = dbView.Name
}
if len(viewPayload.Filters) == 0 {
	viewPayload.Filters = dbView.Filters
}
if viewPayload.Visibility == "" {
	viewPayload.Visibility = dbView.Visibility
}
if len(viewPayload.Metadata) == 0 {
	viewPayload.Metadata = dbView.Metadata
}
```

So the handler's contract is "send what you want to change." The schema's contract is "always send a non-empty name." They disagree, and the schema is the one that is wrong about the operation.

## Why this is not merely cosmetic

There is no valid body for a visibility-only update:

- Omitting `name` violates `required: [name]`.
- Sending `name: ""` violates `minLength: 1`.
- Sending the client's cached `name` is a **write**: if the view was renamed elsewhere after the client read it, the update silently reverts the rename.

Today only the third option is schema-valid, and it is the one that can lose data.

Nothing enforces the schema at runtime right now - `UpdateView` unmarshals by hand and there is no `OapiRequestValidator`/kin-openapi middleware anywhere in `meshery-cloud`, so `name: ""` currently reaches the retain-existing branch and works. That is precisely the fragility: the moment request validation is introduced upstream, every partial view update starts failing with a 400, and the failure will look like a regression in the consumer rather than a latent schema defect.

## Impact

Found during a schema-consumer audit of Kanvas (`layer5labs/meshery-extensions`), which repointed local duplicates of schemas-owned operations to their canonical clients. `updateView` repointed cleanly and the repoint fixed a live bug - the Share modal had been sending `{"body":{"visibility":...}}`, which `MesheryViewPayload` ignored, so toggling a view's visibility was a silent no-op that reported success.

Kanvas's visibility toggle now sends `name: ""` deliberately, to hit the provider's retain-existing path rather than risk clobbering a concurrent rename. That is a knowing deviation from the canonical payload, annotated at `buildViewVisibilityUpdate` with a pointer to this issue and recorded in the repo's deviations table alongside #1140, #1141, #1134, #1146 and #708.

## Ask

Give `updateView` its own partial-update payload rather than reusing the create body - e.g. a `ViewUpdatePayload` with no `required` list, every property optional, and `minLength` retained on `name` so a *supplied* name still cannot be empty:

```yaml
    ViewUpdatePayload:
      type: object
      description: >
        Partial update for a view. Every property is optional; omitted properties
        retain their stored value.
      properties:
        name:
          type: string
          maxLength: 255
          minLength: 1
        filters:
          type: object
        visibility:
          type: string
        metadata:
          type: object
```

That makes "change only the visibility" expressible as `{"visibility": "public"}`, which is what the handler has always implemented, and lets Kanvas drop the `name: ""` workaround.

Worth checking the sibling constructs in the same pass: any `PUT`/partial-update operation that `$ref`s a create-shaped `requiredsomething` payload has the same defect. `updateWorkspace`, `updateEnvironment` and `updateOrg` are the obvious candidates.

MU5A:
Fixes #1147

`updateView` is a partial-update PUT but reused the create-shaped `ViewPayload`, which requires a non-empty `name` on every request. There was no schema-valid way to express a visibility-only update: omitting `name` violates the `required` list, an empty `name` violates `minLength`, and resending the client's cached `name` is a write that can silently revert a rename made elsewhere.

The provider (`meshery-cloud`'s `UpdateView`) has always implemented a real partial update, retaining any field the caller omits. The schema's "always send a non-empty name" contract disagreed with that. Nothing enforces the schema at runtime yet only because `UpdateView` unmarshals by hand with no request-validation middleware in front of it — the issue notes this is exactly the kind of latent defect that turns into a live 400 regression the moment validation is introduced upstream.

Per the issue's own "Ask": added `ViewUpdatePayload`, with every property optional and `minLength` retained on a *supplied* name, and pointed `updateView`'s requestBody at it instead of the create payload — in both `v1beta1` and `v1beta2` (the issue notes the defect exists in both). Also corrected `ViewPayload`'s now-inaccurate "creating or updating" description since it's create-only now.

Regenerated the Go models, TypeScript types, and RTK client so everything stays in sync.

## Test plan
- [x] `make generate-golang` — confirmed `ViewUpdatePayload.Name` generates as `*string` (optional) vs. `ViewPayload.Name string` (required)
- [x] `make generate-rtk` and `make generate-ts` — confirmed `updateView`'s operation-level requestBody now has `name?: string` instead of `name: string`
- [x] `test-gofmt` and `test-rtk` regression suites pass

Note: while verifying generation locally I hit a pre-existing `SyntaxError` in `build/generate-golang.js` (duplicate `const helperPath`, introduced by #1180) that breaks `make generate-golang` on current master for everyone. That's unrelated to this fix, so I've filed it separately rather than bundling it in here.

## Summary by CodeRabbit

* **New Features**
  * Added support for partial view updates, allowing clients to provide only the fields they want to change.
  * Omitted update fields now retain their existing values.
  * View names remain subject to non-empty validation when supplied.

* **Documentation**
  * Clarified that view creation payloads are intended for creating views only.
  * Updated API and client type definitions to reflect optional fields for view updates.

### DIFF
--- a/models/v1beta1/view/view.go
+++ b/models/v1beta1/view/view.go
@@ -113,7 +113,7 @@ type MesheryViewWithLocation struct {
 	WorkspaceName string `db:"workspace_name" json:"workspace_name,omitempty" yaml:"workspace_name,omitempty"`
 }
 
-// ViewPayload Payload for creating or updating a view.
+// ViewPayload Payload for creating a view.
 type ViewPayload struct {
 	// Filters Filter configuration for this view.
 	Filters core.Map `json:"filters,omitempty" yaml:"filters,omitempty"`
@@ -123,6 +123,21 @@ type ViewPayload struct {
 
 	// Name Display name of the view.
 	Name string `json:"name,omitempty" yaml:"name,omitempty"`
+
+	// Visibility Visibility level of the view.
+	Visibility *string `json:"visibility,omitempty" yaml:"visibility,omitempty"`
+}
+
+// ViewUpdatePayload Partial update for a view. Every property is optional; an omitted property retains its stored value. A supplied name still cannot be empty.
+type ViewUpdatePayload struct {
+	// Filters Filter configuration for this view.
+	Filters core.Map `json:"filters,omitempty" yaml:"filters,omitempty"`
+
+	// Metadata Metadata associated with the view.
+	Metadata core.Map `json:"metadata,omitempty" yaml:"metadata,omitempty"`
+
+	// Name Display name of the view.
+	Name *string `json:"name,omitempty" yaml:"name,omitempty"`
 
 	// Visibility Visibility level of the view.
 	Visibility *string `json:"visibility,omitempty" yaml:"visibility,omitempty"`
--- a/models/v1beta2/view/view.go
+++ b/models/v1beta2/view/view.go
@@ -124,7 +124,7 @@ type MesheryViewWithLocation struct {
 	WorkspaceName string `db:"workspace_name" json:"workspaceName,omitempty" yaml:"workspaceName,omitempty"`
 }
 
-// ViewPayload Payload for creating or updating a view.
+// ViewPayload Payload for creating a view.
 type ViewPayload struct {
 	// Filters Filter configuration for this view.
 	Filters core.Map `json:"filters,omitempty" yaml:"filters,omitempty"`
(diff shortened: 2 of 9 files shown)


Junnygram (approved):

emphor11 (approved):
Reviewed the schema change across v1beta1/v1beta2, the generated Go/TypeScript models, and the RTK client.

The update contract now correctly allows omitted fields while retaining `minLength: 1` for a supplied name, and `updateView` references the dedicated `ViewUpdatePayload` in both API versions. The creation payload remains strict.

I don't see a blocking correctness issue in the current implementation.

PR #1189, opened 2026-09-05
Give updateEnvironment its own partial-update payload

MU5A:
This is the sibling defect #1147 itself flagged as an obvious next place to check: "updateWorkspace, updateEnvironment and updateOrg are the obvious candidates" for the same "PUT reuses a create-shaped required-something payload" pattern.

Checked all three against current master before writing anything:
- `updateWorkspace` already has its own `WorkspaceUpdatePayload` (only `organizationId` required there, not `name`). Not affected.
- `updateOrg`'s `OrganizationPayload` has no `required` list at all. Not affected.
- `updateEnvironment` (both v1beta1 and v1beta3) reuses the create-shaped `EnvironmentPayload`, which requires both `name` and `organizationId`/`organization_id` on every request. **Affected.**

`organizationId`'s own description states it "cannot be changed after creation," so requiring it on update forces a client to resend an immutable value just to change anything else. There was also no schema-valid way to send a description-only update without resending `name` too — a write that can silently revert a concurrent rename, the same risk class already fixed for `updateView` in #1147.

## What I changed

Added `EnvironmentUpdatePayload` (every property optional, `organizationId` accepted but not required) in both `v1beta1` and `v1beta3`, and pointed `updateEnvironment`'s requestBody at it instead of the create payload. Kept `organizationId` in the update payload for now, since I can't verify from this repo alone whether the private handler does anything with it on update — whether it should be dropped entirely given it's immutable is a separate design question I'm leaving to maintainers rather than deciding unilaterally.

Regenerated the Go models, TypeScript types, and RTK client so nothing is left stale.

## Test plan
- [x] `make generate-golang`, `make generate-ts`, `make generate-rtk` — confirmed `EnvironmentUpdatePayload.Name` generates as optional vs. `EnvironmentPayload.Name` required
- [x] Existing gofmt/RTK regression suites pass

Unlike #1147, I don't have a confirmed live incident for this one (no equivalent to the Kanvas Share-modal bug that motivated #1147) — flagging that distinction honestly since it's asked for in review.

## Summary by CodeRabbit

* **New Features**
  * Environment updates now support partial changes, allowing users to modify only the name and/or description.
  * Organization identifiers are accepted during updates but remain unchanged after environment creation.
  * Updated API definitions and client types consistently support the new partial-update behavior.

### DIFF
--- a/models/v1beta1/environment/environment.go
+++ b/models/v1beta1/environment/environment.go
@@ -89,6 +89,15 @@ type EnvironmentPayload struct {
 	OrgId uuid.UUID `json:"organization_id" yaml:"organization_id"`
 }
 
+// EnvironmentUpdatePayload Partial update for an environment. Every property is optional; an omitted property retains its stored value. organization_id is accepted for symmetry with the create payload but is not actually mutable - see its own description - so omitting it is the normal case.
+type EnvironmentUpdatePayload struct {
+	Description core.Text `json:"description,omitempty" yaml:"description,omitempty"`
+	Name        core.Text `json:"name,omitempty" yaml:"name,omitempty"`
+
+	// OrgId The organization cannot be changed after creation. Accepted here only so a client that already has this value does not have to strip it before sending a partial update.
+	OrgId uuid.UUID `json:"organization_id" yaml:"organization_id"`
+}
+
 // EnvironmentId defines model for environmentId.
 type EnvironmentId = core.Id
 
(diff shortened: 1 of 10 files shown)


Junnygram:
@MU5A ,It's advised to respond to each comment with either accepting or rejecting, followed by a why. You currently have two open comments, please share your feedback on those. Thanks

MU5A:
@Junnygram both accepted, these were real. Sorry for the delay in getting to them.

CodeRabbit caught a bug I introduced myself: I copied x-go-type-skip-optional-pointer: true onto organizationId in EnvironmentUpdatePayload directly from the create payload, where it's correct because that field is required there. In the update payload it's optional, and Go's omitempty does not omit a zero-valued fixed-size uuid.UUID array, so an omitted organizationId and an all-zeros one were indistinguishable on the wire, exactly the kind of bug this PR exists to prevent, just on the one field I didn't scrutinize as carefully as the others.

Fixed in both v1beta1 and v1beta3: removed the skip-pointer override, added omitempty to the JSON tag, regenerated Go/TS, confirmed the field is now *uuid.UUID. Validator and regression suites pass.

PR #1193, opened 2026-09-06
fix(v1beta3/academy): drop omitempty from required registrationCount

### LINKED ISSUE #1160: v1beta3 academy registrationCount is required but generated with omitempty, so a genuine 0 is omitted (and it is float32)
## Summary

`registrationCount` on the v1beta3 academy curricula response is declared `required`, then generated with `omitempty` - so a genuine count of **0** is omitted from the payload entirely, making it indistinguishable from an absent key and putting the response in violation of the schema's own `required` constraint. It is also typed `number`, which generates `float32` for what is a count of rows.

## Evidence

`schemas/constructs/v1beta3/academy/api.yml` (v1.3.44) declares it required and then authors the tag itself:

```yaml
    SingleAcademyCurriculaResponse:
      type: object
      allOf:
      - $ref: '#/components/schemas/AcademyCurricula'
      - type: object
        required:
        - registrationCount
        properties:
          registrationCount:
            type: number
            description: Number of registrations associated with this curriculum.
            minimum: 0
            x-oapi-codegen-extra-tags:
              db: registration_count,omitempty
              json: registrationCount,omitempty
```

Generated Go (`models/v1beta3/academy/academy.go:153`):

```go
RegistrationCount float32 `db:"registration_count,omitempty" json:"registrationCount,omitempty" yaml:"registrationCount,omitempty"`
```

So the `omitempty` is not a generator default that could be argued about - it is written into the schema, on a field the same file marks required, one key above.

## Consequences

1. **A zero is unrepresentable.** `encoding/json` drops a `0` under `omitempty`, so a curriculum with no registrations serializes with no `registrationCount` at all. A consumer cannot distinguish "measured zero" from "field missing", which is the difference between "nobody enrolled" and "we could not read it".
2. **The response violates its own contract.** Any zero-registration curriculum emits a document that fails validation against the `required` list.
3. **`float32` for a row count.** `type: number` generates a float where an integer is meant; `minimum: 0` is already declared, so the intent is a non-negative count. Beyond being wrong-typed, float32 loses exactness above 2^24.

## Asks

1. Drop `omitempty` from the `json` (and `db`) extra-tags for a required field, or drop it from `required` - the two cannot both be right.
2. Type it as an integer (`type: integer`, `format: int64` or similar) rather than `number`.

The `db` tag deserves the same look: `omitempty` there affects ORM writes, and a zero count that silently does not persist has the same shape of bug on the storage side.

## Context

Found while fixing blank date columns in layer5io/meshery-cloud#5985. The instructor console's registration counts have a history with this construct - a stale snake_case read of `registration_count` after the v1beta3 migration made the overview report "0 registrations" while the learners tab listed hundreds - and this defect makes a real zero look identical to that failure mode, so it is worth closing before the next such bug is diagnosed.

akkki007:
## Description

`registrationCount` is declared `required` on the v1beta3 academy curricula responses, but the schema also authors `json: registrationCount,omitempty` through `x-oapi-codegen-extra-tags`. A genuine count of **0** was therefore omitted from the payload entirely — indistinguishable from an absent key, and in violation of the schema's own `required` constraint.

This removes `omitempty` from both the `db` and `json` tags, and changes `type: number` to `type: integer`.

Fixes #1160

## Notes for reviewers

**The defect occurs twice.** The issue cites `SingleAcademyCurriculaResponse` (`api.yml:863-871`); `AcademyCurriculaWithMetrics` (`api.yml:946-954`) carries an identical block. Both are fixed here — `models/v1beta3/academy/academy.go:153` and `:693`.

**On `number` → `integer` (`float32` → `int`).** This is the part worth a second opinion. The justification is internal precedent: every other count in this construct is already an integer type — `Total int` (`:118`, `:174`), `TotalCount int64` (`:405`), `TotalCount *int` (`:422`, `:444`). `RegistrationCount float32` was the sole outlier for a `minimum: 0` count of rows. I found no consumers of the field in `meshery/meshery`, but **`meshery-cloud` and `meshery-extensions` are private and I could not check them** — so the `consumer-audit` job may have more to say than I can see. Happy to split the type change into a separate PR if you'd rather land the `omitempty` fix on its own.

**Regeneration.** `models/v1beta3/academy/academy.go` was regenerated with `make generate-golang`. One thing to flag: this had to be done with `oapi-codegen` **v2.5.1**, which is what the committed headers declare — not the **v2.5.0** pinned in `go.mod`. At the pinned version the tool does not build against the pinned `kin-openapi`:

```
pkg/codegen/schema.go:899:13: invalid operation: v == element.Ref
  (mismatched types openapi3.MappingRef and string)
```

and regenerating with it rewrites 77 files with a downgraded version header. That's outside the scope of this PR; I'll open a separate issue for it.

## Verification

All blocking checks from `.github/workflows/schema-audit.yml`, run locally:

```
make validate-schemas                  ✓ no blocking violations
go test ./validation/... ./models/...  ✓ all pass
make test-rtk                          ✓ 2/2 and 3/3
go build ./...                         ✓
gofmt -l models/                       ✓ clean
make generate-ts                       ✓ no committed-file changes
```

## Diff

```diff
-            type: number
+            type: integer
             description: Number of registrations associated with this curriculum.
             minimum: 0
             x-oapi-codegen-extra-tags:
-              db: registration_count,omitempty
-              json: registrationCount,omitempty
+              db: registration_count
+              json: registrationCount
```

```diff
-RegistrationCount float32 `db:"registration_count,omitempty" json:"registrationCount,omitempty" yaml:"registrationCount,omitempty"`
+RegistrationCount int `db:"registration_count" json:"registrationCount" yaml:"registrationCount"`
```

Signed-off-by: Akshay Nazare <akshaynazare3@gmail.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Registration counts are now represented as whole numbers in academy curriculum responses.
  * Registration count fields are consistently included in returned data, including when the value is zero.

### DIFF
--- a/models/v1beta3/academy/academy.go
+++ b/models/v1beta3/academy/academy.go
@@ -150,7 +150,7 @@ type AcademyCurriculaWithMetrics struct {
 	OrgId AcademyCurriculaOrgId `db:"org_id" json:"orgId" yaml:"orgId"`
 
 	// RegistrationCount Number of registrations associated with this curriculum.
-	RegistrationCount float32 `db:"registration_count,omitempty" json:"registrationCount,omitempty" yaml:"registrationCount,omitempty"`
+	RegistrationCount int `db:"registration_count" json:"registrationCount" yaml:"registrationCount"`
 
 	// Slug slug of the Curricula
 	Slug   string      `json:"slug" yaml:"slug"`
@@ -690,7 +690,7 @@ type SingleAcademyCurriculaResponse struct {
 	OrgId AcademyCurriculaOrgId `db:"org_id" json:"orgId" yaml:"orgId"`
 
 	// RegistrationCount Number of registrations associated with this curriculum.
-	RegistrationCount float32 `db:"registration_count,omitempty" json:"registrationCount,omitempty" yaml:"registrationCount,omitempty"`
+	RegistrationCount int `db:"registration_count" json:"registrationCount" yaml:"registrationCount"`
 
 	// Slug slug of the Curricula
 	Slug   string      `json:"slug" yaml:"slug"`
--- a/schemas/constructs/v1beta3/academy/api.yml
+++ b/schemas/constructs/v1beta3/academy/api.yml
@@ -863,12 +863,12 @@ components:
         - registrationCount
         properties:
           registrationCount:
-            type: number
+            type: integer
             description: Number of registrations associated with this curriculum.
             minimum: 0
             x-oapi-codegen-extra-tags:
-              db: registration_count,omitempty
-              json: registrationCount,omitempty
+              db: registration_count
+              json: registrationCount
           invitation:
             $ref: ../invitation/api.yml#/components/schemas/Invitation
             x-go-type: invitationv1beta3.Invitation
@@ -946,12 +946,12 @@ components:
         - registrationCount
         properties:
           registrationCount:
-            type: number
+            type: integer
             description: Number of registrations associated with this curriculum.
             minimum: 0
             x-oapi-codegen-extra-tags:
-              db: registration_count,omitempty
-              json: registrationCount,omitempty
+              db: registration_count
+              json: registrationCount
     AcademyCurriculaListResponse:
       type: object
       properties:


akkki007:
Closing this as a duplicate — I should have caught it before opening.

**#1170 already does this**, and has since 2026-08-20. It works from #1116, which is the original report; #1160 (what I worked from) is a later duplicate of it, and #1116 is assigned to @dhruveshmishra. I searched open PRs for `#1160` in the body and found nothing, which is exactly why I missed it — the work was being tracked under the other issue number.

For anyone comparing the two: **#1170 is the better change.** It fixes both blocks, and it also sets `format: int64`, which this PR omits — plain `integer` generates Go `int`, whose width is platform-dependent, so `int64` is the right call for a count coming out of a database column.

The only thing here that #1170 doesn't carry is the regenerated `models/v1beta3/academy/academy.go`. That isn't a gap: `generate-artifacts-from-schemas.yml` regenerates and auto-commits artifacts on push to `master`, so merging #1170 produces the same `int64`/no-`omitempty` struct tags without a hand-committed model diff.

Suggest closing #1160 as a duplicate of #1116 as well, so the two don't keep attracting parallel PRs.

Sorry for the extra item in the review queue.

PR #1195, opened 2026-09-06
fix(typescript): give badge and invitation root exports versioned names

### LINKED ISSUE #1154: Root barrel exports `BadgeSchema` and `InvitationSchema` silently pin the retired v1beta1 construct
## Summary

The package root re-exports ~15 OpenAPI construct documents. Almost all of them
carry the construct version in the identifier, so a consumer can see which
edition they are getting at the import site:

```
CategoryDefinitionV1Beta1OpenApiSchema        -> constructs/v1beta1/category/CategorySchema
ComponentDefinitionV1Beta2OpenApiSchema       -> constructs/v1beta2/component/ComponentSchema
DesignDefinitionV1Beta2OpenApiSchema          -> constructs/v1beta2/design/DesignSchema
RelationshipDefinitionV1Alpha3OpenApiSchema   -> constructs/v1alpha3/relationship/RelationshipSchema
InvitationDefinitionV1Beta2OpenApiSchema      -> constructs/v1beta2/invitation/InvitationSchema
...
```

**Exactly two do not**, and both silently resolve to `v1beta1`:

| Root export | Resolves to | Newer construct exists | Go model consumers bind |
| --- | --- | --- | --- |
| `BadgeSchema` | `constructs/v1beta1/badge/BadgeSchema` | `v1beta2` | `models/v1beta2/badge` |
| `InvitationSchema` | `constructs/v1beta1/invitation/InvitationSchema` | `v1beta2`, `v1beta3` | `models/v1beta3/invitation` |

(Verified against `@meshery/schemas@1.3.44` by deep-equality between each root
export and every file under `dist/constructs/`.)

## Why it bites

In both cases the difference between the pinned edition and the current one is
the **snake_case -> camelCase identifier rename**, i.e. exactly the field names
a consumer addresses:

```
badge       v1beta1: id, org_id,   label, name, description, image_url,  created_at, ...
            v1beta2: id, orgId,    label, name, description, imageUrl,   createdAt,  ...

invitation  v1beta1: id, owner_id, is_default, org_id, expires_at, accepted_by, created_at, ...
            v1beta3: id, owner,    isDefault,  orgId,  expiresAt,  acceptedBy,  createdAt,  ...
```

The Go models these constructs generate are what providers actually bind, and
they are camelCase — `models/v1beta2/badge.Badge` is tagged `json:"imageUrl"` /
`json:"orgId"`, `models/v1beta3/invitation.Invitation` is tagged `json:"owner"` /
`json:"expiresAt"`. The generated RTK operations agree (`createOrUpdateBadge`'s
body declares `imageUrl`, `orgId`).

So an unsuffixed export name reads as *the* schema for that construct while
pinning the retired edition, and a consumer that addresses the live camelCase
field names inherits the mismatch **with no error anywhere**: `_.pick` /
property lookups against the document return nothing, and the failure surfaces
only as a form field that never renders.

That is not hypothetical — it shipped:
[layer5io/meshery-cloud#5917](https://github.com/layer5io/meshery-cloud/issues/5917).
The Recognitions create/edit modal projected its RJSF schema out of the root
`BadgeSchema` while addressing `imageUrl`. The image field never rendered, and
because `required` still named `imageUrl`, RJSF refused every submit against a
field the user could not see: "Create Recognition" sent no request at all. The
consumer's field names were correct throughout; only the import source was wrong.

## Suggested fix

Either is fine, and either one makes the mistake impossible rather than merely
reviewable:

1. Rename to match every sibling — `BadgeDefinitionV1Beta2OpenApiSchema`,
   `InvitationDefinitionV1Beta3OpenApiSchema` — keeping the old names as
   deprecated aliases for a release; or
2. Repoint the two unsuffixed exports at the current construct edition.

A packaging test asserting that no root export of a construct document is
version-ambiguous would keep it from recurring as new constructs are added.

## Consumer-side note

`meshery-cloud` has repointed its badge consumer at the
`constructs/v1beta2/badge/BadgeSchema` subpath and added an architecture guard
(`ui/__tests__/architecture/schema-construct-version-parity.test.ts`) that
refuses any import of these two exports and asserts every construct the UI
imports matches the version its Go server binds. `InvitationSchema` is not
consumed from the barrel there — both consumers already use the `v1beta3`
subpath — so it is currently a latent trap rather than a live defect.

akkki007:
## Description

Root exports of construct documents carry the construct version in the identifier, so a consumer can see which edition it is getting at the import site. Exactly two did not — `BadgeSchema` and `InvitationSchema` read as *the* schema for their construct while resolving to the retired v1beta1 edition, which still carries the snake_case identifiers. A consumer addressing the live camelCase names got no error anywhere; property lookups against the wrong document just return nothing.

This adds the correctly-named current editions, names the existing v1beta1 imports explicitly, and keeps the two unsuffixed names as deprecated aliases resolving to exactly what they resolved to before.

Fixes #1154

## Notes for reviewers

**Nothing breaks.** `BadgeSchema` and `InvitationSchema` still export the v1beta1 documents. This is purely additive plus a deprecation marker; the decision about when to drop the aliases is left to you.

**`@deprecated` on an export specifier does not survive the build.** This is the one non-obvious part. `tsup` flattens the export list into a single statement, so a JSDoc attached to a specifier is dropped and never reaches `dist/index.d.ts` — consumers would get no editor signal, which is most of the point. Declaring the aliases as consts emits `declare const BadgeSchema` with the comment intact. Verified in both `dist/index.d.ts` and `dist/index.d.mts`:

```ts
/**
 * @deprecated Version-ambiguous. Resolves to the retired v1beta1 badge
 * construct, which still carries the snake_case identifiers (`image_url`,
 * `org_id`)...
 */
declare const BadgeSchema: Record<string, unknown>;
```

**The packaging guard.** `typescript/index.test.ts` resolves each root export back to the construct path it imports and fails if the exported name doesn't encode that version, with the two aliases explicitly allowlisted. A second assertion pins those aliases to v1beta1, so repointing them cannot happen incidentally — that would be a breaking change and should be deliberate.

I confirmed the guard fails against the pre-fix state rather than passing vacuously:

```
not ok 1 - every root construct-document export names the construct version it resolves to
  BadgeSchema -> ./generated/v1beta1/badge/BadgeSchema (expected the name to contain V1Beta1)
```

**One thing needing your call.** `npm test` only runs in `generate-artifacts-from-schemas.yml`, which triggers on push to master — so this guard will not gate PRs as things stand. I've left the workflows untouched; happy to add a step to the blocking job in `schema-audit.yml` if you want it enforced pre-merge.

## Verification

```
npm test                            ✓ 6 pass, 0 fail (4 before, +2 here)
npm run build                       ✓ builds; @deprecated present in dist/index.d.ts and .d.mts
npx tsc --noEmit                    ✓ no new errors (6 pre-existing node_modules resolution errors, identical on master)
negative test (alias de-allowlisted) ✓ guard fails as intended
```

Signed-off-by: Akshay Nazare <akshaynazare3@gmail.com>

### DIFF
--- a/typescript/index.test.ts
+++ b/typescript/index.test.ts
@@ -1,0 +1,128 @@
+import { test } from "node:test";
+import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+import { join } from "node:path";
+
+/**
+ * Packaging guard for meshery/schemas#1154.
+ *
+ * Root exports of construct documents carry the construct version in the
+ * identifier, so a consumer can see which edition it is getting at the import
+ * site. Exactly two did not: `BadgeSchema` and `InvitationSchema` read as *the*
+ * schema for their construct while resolving to the retired v1beta1 edition.
+ *
+ * That edition still carries the snake_case identifiers (`image_url`, `org_id`,
+ * `owner_id`, `expires_at`). A consumer addressing the live camelCase names got
+ * no error anywhere - property lookups against the wrong document simply return
+ * nothing - so the failure surfaced only as a form field that never rendered
+ * (layer5io/meshery-cloud#5917).
+ *
+ * These assertions fail if a new version-ambiguous root export is introduced.
+ */
+
+const indexSource = readFileSync(join(import.meta.dirname, "index.ts"), "utf-8");
+
+/**
+ * Version-ambiguous root exports retained as deprecated aliases for one
+ * release. Nothing may be added here - entries only leave, when the alias is
+ * dropped.
+ */
+const DEPRECATED_AMBIGUOUS_EXPORTS = new Set(["BadgeSchema", "InvitationSchema"]);
+
+/**
+ * Local binding -> the generated construct path it resolves to, following the
+ * `const alias = binding;` indirection the deprecated aliases are declared with.
+ */
+function importedConstructPaths(): Map<string, string> {
+  const paths = new Map<string, string>();
+  const importLine = /^import\s+(\w+)\s+from\s+"(\.\/generated\/[^"]+)";$/gm;
+  for (const [, binding, path] of indexSource.matchAll(importLine)) {
+    paths.set(binding, path);
+  }
+
+  const constAlias = /^const\s+(\w+)\s*=\s*(\w+);$/gm;
+  for (const [, alias, target] of indexSource.matchAll(constAlias)) {
+    const path = paths.get(target);
+    if (path) paths.set(alias, path);
+  }
+
+  return paths;
+}
+
+/** The `export { ... }` block that publishes the construct documents. */
+function rootSchemaExports(): Array<{ local: string; exported: string }> {
+  const blockStart = indexSource.indexOf("// Export schemas");
+  assert.notEqual(blockStart, -1, "expected an '// Export schemas' block in index.ts");
+
+  const open = indexSource.indexOf("export {", blockStart);
+  const close = indexSource.indexOf("};", open);
+  assert.ok(open !== -1 && close !== -1, "expected a closed export block");
+
+  const body = indexSource
+    .slice(open + "export {".length, close)
+    .replace(/\/\*[\s\S]*?\*\//g, "")
+    .replace(/\/\/[^\n]*/g, "");
+
+  return body
+    .split(",")
+    .map((entry) => entry.trim())
+    .filter(Boolean)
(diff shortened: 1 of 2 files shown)


akkki007:
Hii @ritzorama 
Can you review this pull request ?
Let me know if any shape changes that you want

Issue #1194, opened 2026-09-06
oapi-codegen tool pin (v2.5.0) does not build and cannot reproduce the committed models (generated by v2.5.1)


akkki007:
## Summary

`go.mod` pins `oapi-codegen` at **v2.5.0** and exposes it through a `tool` directive, but every committed file under `models/` declares it was generated by **v2.5.1**. The pinned version cannot reproduce the committed output — and in fact does not build.

## The tool does not compile at the pinned version

```console
$ go tool oapi-codegen --version
# github.com/oapi-codegen/oapi-codegen/v2/pkg/codegen
../../../go/pkg/mod/github.com/oapi-codegen/oapi-codegen/v2@v2.5.0/pkg/codegen/schema.go:899:13:
  invalid operation: v == element.Ref (mismatched types openapi3.MappingRef and string)
```

v2.5.0 is incompatible with the `kin-openapi` version resolved by this module (`v0.149.0`), which changed `Discriminator.Mapping` element typing. So `make generate-golang` cannot be run via the repo's own tool directive at all.

## Installing v2.5.0 separately reproduces the wrong output

Following the `go.mod` pin explicitly and regenerating rewrites **77 files**, each with a downgraded header:

```diff
-// Code generated by github.com/oapi-codegen/oapi-codegen/v2 version v2.5.1 DO NOT EDIT.
+// Code generated by github.com/oapi-codegen/oapi-codegen/v2 version v2.5.0 DO NOT EDIT.
```

Only v2.5.1 reproduces the tree as committed.

## Impact

A contributor regenerating models — which `make generate-golang` requires for any schema change — either hits a build failure via `go tool`, or silently produces a 77-file diff if they install the pinned version by hand. Both send the wrong signal about what their change actually did. I hit this on #1160 / #1193 and had to determine the real version from the generated headers.

## Suggested fix

Bump the pin to match the version that produced the tree:

```diff
-	github.com/oapi-codegen/oapi-codegen/v2 v2.5.0 // indirect
+	github.com/oapi-codegen/oapi-codegen/v2 v2.5.1 // indirect
```

with the corresponding `go.sum` update. Worth considering a CI step that regenerates and fails on a dirty tree, which would have caught the drift when it appeared.

## Environment

- `meshery/schemas` @ `5ee8de9c`
- Go 1.26.4 (via `GOTOOLCHAIN=auto`)
- `kin-openapi v0.149.0` per `go.mod`

akkki007:
[...]

PR #1197, opened 2026-09-07
docs(openapi): document credential_id and selectedOrganizationID as deprecated aliases

### LINKED ISSUE #1136: Document credential_id and selectedOrganizationID as deprecated aliases in the OpenAPI
## Summary

Meshery Server now accepts two legacy wire spellings as deprecated aliases, but neither is described in the OpenAPI. Since `docs/data/openapi.yml` is generated here and copied into `meshery/meshery` by its `[Docs] Update REST API Documentation` job, the meshery repo cannot document them itself - the description has to originate in schemas.

Raised by the docs review on meshery/meshery#PR_PLACEHOLDER, which introduced the aliases.

## The two aliases

**1. `credential_id` on `DELETE /api/integrations/credentials`**

The canonical param is `credentialId` (what the generated `deleteUserCredential` operation already sends). Meshery Server previously read *only* `credential_id`, so the generated client resolved a nil UUID and deleted nothing. The server now reads `credentialId` first and falls back to `credential_id`, so older clients keep working.

**2. `selectedOrganizationID` in the `Preference` body (`POST /api/user/prefs`)**

The canonical property is `selectedOrganizationId` (schemas `v1beta1` `user.Preference`, and what meshery-cloud reads and writes). Meshery Server previously emitted and read the all-caps `selectedOrganizationID`, so the selected organization never round-tripped through a remote provider. It now emits the canonical spelling and accepts the legacy one on read, so preferences already persisted under the old key survive.

## Ask

Mark both as deprecated aliases in the OpenAPI, matching the existing pattern in this repo. There is already precedent on the very same credentials path - `docs/data/openapi.yml` (generated from here) carries:

```yaml
page_size:
  type: integer
  description: Deprecated snake_case alias of `pageSize`.
  minimum: 0
  deprecated: true
total_count:
  type: integer
  description: Deprecated snake_case alias of `totalCount`.
  minimum: 0
  deprecated: true
```

So:

- `DELETE /api/integrations/credentials` - add a `credential_id` query parameter with `deprecated: true` and a description naming `credentialId` as the canonical spelling.
- `Preference` - add a `selectedOrganizationID` property with `deprecated: true` and a description naming `selectedOrganizationId` as the canonical spelling.

## Why it matters

Publishing them is what makes the fallbacks removable later. Right now the only record that these aliases are accepted lives in Go comments in `meshery/meshery`, so a future contributor reading the OpenAPI has no way to know the fallback exists, and no way to judge when it is safe to drop. Documenting them as deprecated gives clients a migration signal and gives us a defensible point to delete the fallbacks once no client sends them.

akkki007:
## Description

Fixes #1136

Meshery Server accepts two legacy wire spellings that the OpenAPI does not describe. Because `docs/data/openapi.yml` is generated here and copied into `meshery/meshery` by its `[Docs] Update REST API Documentation` job, the description has to originate in this repo.

**The issue's `meshery/meshery#PR_PLACEHOLDER` resolves to meshery/meshery#21191** ("[Server][UI] Repoint schemas-owned contracts at `@meshery/schemas`", merged 2026-08-07, commit `baa98b5`). Both fallbacks landed there and are on `master` today:

| Alias | Server code | Behavior |
|---|---|---|
| `credential_id` | `server/handlers/credentials_handlers.go:150-153` | reads `credentialId`, falls back to `credential_id` |
| `selectedOrganizationID` | `server/models/preference.go:76-95` | custom `UnmarshalJSON` accepting both spellings |

### `credential_id` — a real deprecated parameter

Added to `v1beta2/credential/api.yml` only, since `v1beta1/credential` carries `x-deprecated: true`. It follows the `pagesize` parameter already sitting in the same `components/parameters` block: `deprecated: true` plus a description naming the canonical spelling.

The canonical `credentialIdQuery` stays `required: true` and the alias is optional, so a generated client never sends both.

### `selectedOrganizationID` — described, *not* declared

My first push declared this as its own `deprecated: true` property on `Preference`. **That was wrong, and it would have caused the exact bug the alias exists to prevent.** I caught it self-reviewing and have reverted it.

`Preference.Scan` (`models/*/user/helpers.go`) routes through `core.MapToStruct` → `encoding/json`, whose field matching **is case-insensitive**. So a legacy row stored under the all-caps key already resolves onto `SelectedOrganizationId` on its own, with no schema change at all:

```
{"selectedOrganizationID":"3f2504e0-4f89-11d3-9a0c-0305e82c3301"}

  on master        → SelectedOrganizationId = 3f2504e0-4f89-11d3-9a0c-0305e82c3301  ✅
  with the property → SelectedOrganizationId = 00000000-0000-0000-0000-000000000000  ❌
```

Declaring the property generates a second field whose tag matches the legacy key **exactly**, and `encoding/json` prefers an exact match over a case-insensitive one. The value lands on the alias field and the canonical field stays zero — silently un-selecting the organization for precisely the rows the alias was added to rescue. No error, no failing type check.

So the alias is documented in the canonical property's `description` instead, and `TestPreferenceScan_LegacyAllCapsKeyResolvesToCanonicalField` pins the behavior in v1beta1, v1beta2 and v1beta3 so nobody re-adds the property later.

## A separate pre-existing bug this turned up

Not fixed here — flagging it because it is in the same code path and I would rather not bundle it.

`Preference.Scan` deliberately sanitizes junk out of `selectedOrganizationId`, because legacy rows hold empty strings there and `MapToStruct` would otherwise fail and block callers like OAuth sign-in. The guard deletes only the exact lowercase key:

```go
if v, ok := mapVal["selectedOrganizationId"]; ok { … }
```

But since matching is case-insensitive, a legacy row under the **all-caps** key reaches `MapToStruct` unsanitized. On current `master`:

```
{"selectedOrganizationID":""}  → uuid: incorrect UUID length 0 in string ""
```

That is the same sign-in-blocking failure the guard was written to prevent, still open for the all-caps population — which is exactly the population meshery/meshery#21191 added the fallback for. Happy to open an issue, or fold the fix in here if you would rather have it in one PR.

## Verification

Everything the `Schema Audit` blocking job runs, locally:

| Check | Result |
|---|---|
| `make validate-schemas` | ✅ no blocking violations |
| `go test ./validation/... ./models/...` | ✅ pass |
| `make test-rtk` | ✅ pass |
| `make test-ts` | ✅ 4/4 |
| `make consumer-audit` | ✅ exit 0 |

I also diffed the advisory audit across the change: `make audit-schemas-full` emits **2014 lines before and 2014 after**, with no new findings — the only textual difference is map-iteration ordering.

Per `AGENTS.md` ("do not manually commit regenerated output in normal PRs"), the regenerated `models/` and `typescript/rtk/` output is **not** included; the automation on `master` commits it. I did regenerate locally with `oapi-codegen v2.5.1` — matching what `generate-artifacts-from-schemas.yml` installs — to confirm the output is what I expect before dropping it. (`go.mod` still pins `v2.5.0`; that drift is #1194 and deliberately untouched here.)

## Summary by CodeRabbit

* **Compatibility**
  * Credential deletion now accepts the deprecated `credential_id` query parameter as a legacy alias for `credentialId`. Use `credentialId` for new integrations; both parameters cannot be provided together.
  * User preferences continue to recognize the legacy `selectedOrganizationID` spelling when reading existing data.
  * Use `selectedOrganizationId` when reading and writing preferences. The legacy spelling is deprecated and should not be sent alongside the canonical property.

### DIFF
--- a/models/v1beta1/user/helpers_test.go
+++ b/models/v1beta1/user/helpers_test.go
@@ -36,3 +36,26 @@ func TestPreferenceScan_ValidSelectedOrganizationIdPreserved(t *testing.T) {
 		t.Fatalf("expected %s, got %s", id, p.SelectedOrganizationId)
 	}
 }
+
+// Meshery Server accepts the deprecated all-caps `selectedOrganizationID` on
+// read (meshery/meshery#21191, server/models/preference.go), because rows
+// persisted under that key exist in the wild. core.MapToStruct goes through
+// encoding/json, whose field matching is case-insensitive, so such a row
+// resolves onto the canonical field on its own.
+//
+// That only holds while the alias is NOT declared as its own property. Adding
+// a `selectedOrganizationID` property to the OpenAPI generates a second field
+// whose tag matches the legacy key exactly; encoding/json prefers the exact
+// match, so the value would land on the alias and leave this field zero -
+// silently un-selecting the organization for exactly the rows the alias exists
+// to rescue. See meshery/schemas#1136.
+func TestPreferenceScan_LegacyAllCapsKeyResolvesToCanonicalField(t *testing.T) {
+	id := uuid.Must(uuid.NewV4())
+	p := &Preference{}
+	if err := p.Scan([]byte(`{"selectedOrganizationID":"` + id.String() + `"}`)); err != nil {
+		t.Fatalf("scan with legacy all-caps key should not error, got: %v", err)
+	}
+	if p.SelectedOrganizationId != id {
+		t.Fatalf("legacy all-caps key must resolve onto the canonical field: expected %s, got %s", id, p.SelectedOrganizationId)
+	}
+}
--- a/models/v1beta2/user/helpers_test.go
+++ b/models/v1beta2/user/helpers_test.go
@@ -36,3 +36,26 @@ func TestPreferenceScan_ValidSelectedOrganizationIdPreserved(t *testing.T) {
 		t.Fatalf("expected %s, got %s", id, p.SelectedOrganizationId)
 	}
 }
+
+// Meshery Server accepts the deprecated all-caps `selectedOrganizationID` on
+// read (meshery/meshery#21191, server/models/preference.go), because rows
+// persisted under that key exist in the wild. core.MapToStruct goes through
+// encoding/json, whose field matching is case-insensitive, so such a row
+// resolves onto the canonical field on its own.
+//
+// That only holds while the alias is NOT declared as its own property. Adding
+// a `selectedOrganizationID` property to the OpenAPI generates a second field
+// whose tag matches the legacy key exactly; encoding/json prefers the exact
+// match, so the value would land on the alias and leave this field zero -
+// silently un-selecting the organization for exactly the rows the alias exists
+// to rescue. See meshery/schemas#1136.
+func TestPreferenceScan_LegacyAllCapsKeyResolvesToCanonicalField(t *testing.T) {
+	id := uuid.Must(uuid.NewV4())
+	p := &Preference{}
+	if err := p.Scan([]byte(`{"selectedOrganizationID":"` + id.String() + `"}`)); err != nil {
+		t.Fatalf("scan with legacy all-caps key should not error, got: %v", err)
+	}
+	if p.SelectedOrganizationId != id {
+		t.Fatalf("legacy all-caps key must resolve onto the canonical field: expected %s, got %s", id, p.SelectedOrganizationId)
+	}
+}
--- a/models/v1beta3/user/helpers_test.go
+++ b/models/v1beta3/user/helpers_test.go
@@ -36,3 +36,26 @@ func TestPreferenceScan_ValidSelectedOrganizationIdPreserved(t *testing.T) {
 		t.Fatalf("expected %s, got %s", id, p.SelectedOrganizationId)
 	}
 }
+
+// Meshery Server accepts the deprecated all-caps `selectedOrganizationID` on
+// read (meshery/meshery#21191, server/models/preference.go), because rows
+// persisted under that key exist in the wild. core.MapToStruct goes through
+// encoding/json, whose field matching is case-insensitive, so such a row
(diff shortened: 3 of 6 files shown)


akkki007:
Hii @leecalcote 
Self-review turned up a defect in my own first push, and I have force-pushed a corrected version. Flagging it rather than quietly fixing it, since the diff changed shape.

**What was wrong:** I had declared `selectedOrganizationID` as its own `deprecated: true` property on `Preference`. That would have caused the exact bug the alias exists to prevent.

`Preference.Scan` routes through `core.MapToStruct` → `encoding/json`, whose field matching is case-insensitive — so a legacy all-caps row already resolves onto `SelectedOrganizationId` with no schema change. Declaring the property generates a second field whose tag matches that key *exactly*, and `encoding/json` prefers the exact match:

```
{"selectedOrganizationID":"3f2504e0-4f89-11d3-9a0c-0305e82c3301"}

  master             → SelectedOrganizationId = 3f2504e0-…-0305e82c3301
  with the property  → SelectedOrganizationId = 00000000-…-000000000000
```

Silently un-selecting the organization for precisely the rows the alias was added to rescue. This is the concrete answer to the open question I raised on #1136 — the property half has to be documentation-only, so it is now described in the canonical property's `description`, with a regression test pinning the behaviour in all three live versions.

**Two other corrections in the same push, both from `AGENTS.md`:**

- Dropped the regenerated `models/` and `typescript/rtk/` output — "do not manually commit regenerated output in normal PRs". The diff is now schema plus tests only.
- Removed the AI-attribution trailers from the commit and the PR description, per "No AI attribution in commits, PR descriptions, comments, or code". Apologies for those being there on the first push; I had them on a general default and had not reconciled it against this repo's policy.

The PR description also now records a **pre-existing** bug this turned up in the same code path — `Preference.Scan`'s sanitising guard matches only the exact lowercase key, so a legacy all-caps row holding an empty string still errors on current `master` and blocks sign-in. Not touched here; say the word and I will either open an issue or fold it in.

PR #1210, opened 2026-09-08
Re-export the organization_smtp forms from the package root

### LINKED ISSUE #1207: Publish the RJSF form specs so consumers can read per-field form help from the schema
## Summary

`@meshery/schemas` does not publish the RJSF form specs, so a consumer cannot read per-field form help from the schema and has to transcribe it.

Verified against **1.3.53** (and unchanged in 1.3.54): the package's `exports` map is `.`, `./cloudApi`, `./mesheryApi`, `./api`, `./permissions`, `./constructs/*`, and `schemas/constructs/v1beta1/organization_smtp/forms/configure.json` is absent from `dist` entirely. The Go module ships it, so Go consumers are fine; TypeScript consumers are not.

## Why it matters

That file already carries a per-field `description` written specifically for form help - shorter and more user-facing than the API construct's descriptions. It is exactly what a form should render beside each input.

Because it is unreachable from the package, `layer5io/meshery-cloud` had to keep a **local transcription** of those strings plus **two drift guards** (a Jest check and a Go pin that parses the TypeScript module) to detect when the schema and the copy diverge. That is a local copy of schema-owned content, which the ecosystem's own rules discourage, and it exists solely because the data cannot be imported.

The consuming code, its documentation, and the Go pin all say the transcription and both guards should be deleted once the export lands. Without this issue there is nothing to watch, so the workaround becomes permanent by default.

## Ask

Export the `forms/` specs from the package - either by including them in `dist` under the existing `./constructs/*` subpath, or by adding a documented subpath such as `./constructs/*/forms/*`.

Worth checking in the same pass which other constructs ship `forms/` specs and are equally unreachable, since the same workaround will be needed for each.

## Follow-up in the consumer

Once published, `layer5io/meshery-cloud` deletes the transcription, the Jest drift check and the Go pin, and imports the spec directly.

MU5A:
Fixes #1207

## Correction, read this first

The original version of this PR added a build step that copied raw `forms/*.json` files into `dist/` and a matching exports entry. Aabid Sofi pointed out on Slack that the RJSF forms are already packaged as TypeScript, via `typescript/forms/index.ts` re-exported from `typescript/index.ts`, the file tsup actually bundles into `dist/index.js`/`dist/index.mjs`, the package's root `.` export.

That's correct, and I should have found it before writing any code. Checked it properly once flagged: built the package from a clean `upstream/master` worktree and required the compiled `dist/index.js`/`.mjs` directly, not just read the source and assumed. 11 of the 12 forms resolve exactly as Aabid described. `organization_smtp`'s two forms do not: they're imported and exported from `typescript/forms/index.ts` (added in the same commit that added the construct), but that commit never added them to the separate, manually maintained re-export list in `typescript/index.ts`. So they never reach the published package, which is the actual gap this issue is about, one construct wide rather than twelve.

Dropped the copy-forms.js approach entirely. It worked, but it would duplicate data the package already ships through the typed barrel, and isn't the pattern this codebase uses for every other construct's forms.

## What changed now

- `typescript/index.ts`: added the two missing re-exports, `OrganizationSmtpConfigureRjsfSchemaV1Beta1` and `OrganizationSmtpConfigureRjsfUiSchemaV1Beta1`, in the same alphabetical spot the existing list already keeps.
- `validation/forms_test.go`: added `TestFormsReExportedFromPackageRoot`. Every existing forms test checks that a JSON file is wired into the `typescript/forms/index.ts` barrel; none of them check that the barrel's exports are still re-exported from `typescript/index.ts`, the file that actually reaches npm consumers. That second, silent step is exactly what dropped organization_smtp. The new test parses both files and fails if a barrel export is missing from the package-root re-export block.

Still flagged, not fixed here: the 57 `templates/` directories with a similar shape (static JSON, not generated). Left open whether TypeScript consumers need those the way `meshery-cloud` needed `forms/`.

## Test plan

- [x] Reverted the `typescript/index.ts` fix and reran `TestFormsReExportedFromPackageRoot` alone: it fails with the exact organization_smtp names named in the error message. Restored the fix, reran, passes.
- [x] `go test ./validation/...`: full suite passes.
- [x] `node --test typescript/**/*.test.ts`: passes.
- [x] Built with `tsup`, then required `dist/index.js` and imported `dist/index.mjs` directly in a plain Node script: `OrganizationSmtpConfigureRjsfSchemaV1Beta1` and its UI counterpart resolve in both, and carry the real per-field description text.
- [x] `go run ./cmd/validate-schemas`: no blocking violations.

## Summary by CodeRabbit

* **New Features**
  * Organization SMTP configuration form and UI schemas are now available through the package’s public exports for form integrations.

* **Tests**
  * Added validation to ensure form schemas and UI schemas are correctly re-exported from the package root.


MU5A:
Good catch, confirmed real: the regex was matching the export name anywhere in the isolated block text, including inside a `//` comment, so a commented-out export line would have silently passed the guard instead of failing it.

Fixed in 887204e: strip `//` line comments from the block before matching. Verified both directions locally: commented out `OrganizationSmtpConfigureRjsfSchemaV1Beta1,` and confirmed the test now fails with the expected message, then restored it and confirmed it passes again.

MU5A:
Both real, confirmed. Fixed in 7761f60:

Rewrote the matching entirely instead of patching the regex again: strip both `//` and `/* */` comments, split the block into comma-separated specifiers, and resolve each to the name a consumer actually imports (the alias half of `Name as Alias`, or the bare name), then check for an exact set match rather than substring containment.

While verifying this I found the block-isolation regex itself had the same class of bug: it used a DOTALL `.*?` with no exclusion on braces, so it could span across the unrelated `export { ... }` block and several `export namespace X { ... }` blocks that come before the real forms re-export in this file, pulling all of that into the "block" the specifier parser then choked on. Excluded `{`/`}` from the capture so it can't cross those.

Verified four ways this time: the target export missing entirely, commented out with `//`, commented out with `/* */`, and aliased to a different name. All four now fail with the expected message; restored, the full suite passes.

PR #1213, opened 2026-09-11
Declare 403 for environment purpose refusals

### LINKED ISSUE #1190: Environment/Connection api.yml: declare 403 for deleteEnvironment, addConnectionToEnvironment and removeConnectionFromEnvironment
### Summary

`layer5io/meshery-cloud` now answers **403 Forbidden** for a deliberate policy refusal on three environment operations, but `schemas/constructs/v1beta3/environment/api.yml` documents only 400/401/404 for them. Generated consumers therefore have no knowledge of a status the server really returns.

Affected operations:

- `deleteEnvironment`
- `assignConnectionToEnvironment`
- `removeConnectionFromEnvironment`

### Why the server returns 403

`Environment.purpose` designates an environment the platform provisions to hold organization-level configuration, and resolvers of that configuration trust it. Cloud refuses the three generic verbs above against such an environment: the connections of an administrative environment are managed exclusively by endpoints gated on organization administration, so a caller holding only the ordinary environment-management keys must not reach it through the generic route.

That refusal is an authorization decision, not a failure, and not a missing resource — the caller can `GET` the very same environment. 403 is the honest status, and collapsing it into the generic 500 was itself a defect fixed alongside this.

### Requested change

Add a `403` response to those three operations, described as a refusal to operate on an environment whose `purpose` is not `user`. The wording should make clear it is about the target's designation rather than the caller's identity, since the same caller is authorized for the route and for other environments in the same organization.

### Context

Consumers that need it: `layer5io/meshery-cloud` (the producer here) and `meshery/meshery`'s extension client, which proxies these routes.

Raised from the `Environment.purpose` adoption in meshery-cloud, which closed a privilege escalation where a workspace admin could seize an organization's identity provider. The divergence is knowingly documented in that repository's `docs/reference/environment-purpose.md` pending this change, rather than left for a reader to discover.

PARZIVAL7498:
## Summary
- Adds a `403` (`EnvironmentPurposeForbidden`) response to `deleteEnvironment`, `addConnectionToEnvironment`, and `removeConnectionFromEnvironment`
- Documents that the refusal is about the target environment's `purpose` designation (not `user`), not the caller's identity

Fixes #1190

## Test plan
- [ ] Confirm OpenAPI YAML still validates in schema CI
- [ ] Confirm generated consumers surface `403` on the three operations

## Summary by CodeRabbit

- **Documentation**
  - Updated API documentation to describe `403 Forbidden` responses when attempting to add, remove, or delete connections and environments that are not intended for user operations.
  - Added a reusable response description with a plain-text explanation of the restriction.

### DIFF
--- a/schemas/constructs/v1beta3/connection/api.yml
+++ b/schemas/constructs/v1beta3/connection/api.yml
@@ -346,6 +346,8 @@ paths:
           $ref: "#/components/responses/400"
         "401":
           $ref: "#/components/responses/401"
+        "403":
+          $ref: "#/components/responses/EnvironmentPurposeForbidden"
         "404":
           $ref: "#/components/responses/404"
         "500":
@@ -366,6 +368,8 @@ paths:
           description: Connection removed from environment
         "401":
           $ref: "#/components/responses/401"
+        "403":
+          $ref: "#/components/responses/EnvironmentPurposeForbidden"
         "404":
           $ref: "#/components/responses/404"
         "500":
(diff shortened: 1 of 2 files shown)


priyanshuuu777 (approved):
LGTM! I reviewed the changes against #1190. The 403 response is correctly added for deleteEnvironment, addConnectionToEnvironment, and removeConnectionFromEnvironment, and the EnvironmentPurposeForbidden description clearly explains that the restriction is based on the target environment's purpose. The separate success-status discrepancy is also left untouched.

PR #1215, opened 2026-09-14
Model the resource share/revoke contract

### LINKED ISSUE #1144: Resource share/revoke contract (ResourceAccessMappingPayload) is unmodelled; its rename silently no-opped every consumer
## Summary

The resource share / revoke contract - `POST /api/resources/{resourceType}/{resourceId}/share` and its `ResourceAccessMappingPayload` / `Actor` bodies - is not modelled anywhere in `meshery/schemas`. It exists only as Go structs in meshery-cloud:

`server/models/model_resource_access_mapping_payload.go`:

```go
type Actor struct {
    ActorId   core.Uuid `json:"actorId,omitempty"`
    ActorType string    `json:"actorType,omitempty"`
}

type ResourceAccessMappingPayload struct {
    GrantAccess  []Actor `json:"grantAccess,omitempty"`
    RevokeAccess []Actor `json:"revokeAccess,omitempty"`
    NotifyUsers  bool    `json:"notifyUsers,omitempty"`
}
```

`grep -rn "grantAccess" schemas/` in this repo returns nothing.

## Why this one is worth modelling

This endpoint has the worst possible failure mode for an unmodelled contract. `HandleResourceShare` decodes with a strict `json.Unmarshal` into the struct above, whose every field carries `omitempty`, and then answers **200** regardless. An unrecognised key is not an error - it is dropped, and the caller is told the share succeeded.

That is not hypothetical. Sistent's `ShareModal` - which is upstream of Meshery UI, Layer5 Cloud UI and Kanvas - was posting the pre-flip snake_case body:

```jsonc
{ "grant_access": [{ "actor_id": "...", "actor_type": "user" }], "revoke_access": [], "notify_users": true }
```

against a server that has expected `grantAccess` / `actorId` / `actorType` / `notifyUsers` since the Phase 4 camelCase flip merged on 2026-04-28. Every "Design shared with ..." toast since then was reporting success on a request that granted nothing. Nothing on either side could catch it: sistent hands the body to a host-supplied RTK mutator typed `any`, and the server never rejects.

Had this contract been generated from schemas, the rename would have been a compile error in every consumer instead of a silent no-op for three months. Fixed downstream in layer5io/sistent#1780, which keeps a local type with a pointer to this issue.

## Suggested resolution

Add `Actor`, `ResourceAccessMappingPayload` and the share/revoke operations to the appropriate construct (the resource-access mapping is org-scoped and touches design, filter and view resource types, so `v1beta3` alongside the design construct or its own `resource_access` construct both seem defensible), and have meshery-cloud alias the generated Go types the way it now does for `Team`.

MU5A:
Fixes #1144

## Correction, read this first

The first two commits on this PR added a new `v1beta3/resource_access` construct at `POST /api/resources/{resourceType}/{resourceId}/share`, the path issue #1144 itself states. That path is wrong, and I found this out myself while addressing CodeRabbit's cache-invalidation review comment.

This repo already has an incomplete `handleResourceShare` operation at `POST /api/resource/{resourceType}/share/{resourceId}` (note: singular `resource`, `share` before the id), present identically in `v1beta1`, `v1beta2`, and `v1beta3` design, with a bare, untyped placeholder for its request and response. Confirmed the placeholder's path, not the issue's, is the real one: `meshery/meshery`'s `ui/rtk-query/resource.ts` calls `extensions/api/resource/${resourceType}/share/${resourceId}` today, and that file's own code comment describes the exact bug from issue #1144 verbatim ("Sistent builds this payload and it goes over the wire verbatim. The Cloud API reads camelCase and silently drops anything else"), tying it directly to this URL.

Issue #1144's own `grep -rn "grantAccess" schemas/` (used to conclude nothing was modelled) is exactly why this got missed: the existing placeholder never declared that field, it just never had its real types filled in.

Removed the wrong new construct entirely. The rest of this PR now fills in `Actor` and `ResourceAccessMappingPayload` on the real, existing `handleResourceShare` operation, in all three versions, instead.

## What changed

`POST /api/resource/{resourceType}/share/{resourceId}` and its request body existed only as an untyped `additionalProperties: true` placeholder, and separately as Go structs in meshery-cloud, so the two had drifted with no way to catch it. The failure mode is the worst kind for an unmodelled contract: the handler decodes with a strict `json.Unmarshal` into a struct whose every field carries `omitempty` and answers 200 regardless, so a stale key name produces a share that reports success and grants nothing.

That happened for real. `layer5io/sistent`'s ShareModal posted the pre-rename snake_case body for three months after meshery-cloud flipped the contract to camelCase. Verified end to end in `layer5io/sistent#1786` by replaying both the before and after request bodies through the server's own decode struct.

Added `Actor` and filled in `ResourceAccessMappingPayload` (replacing the untyped placeholder) on the existing operation, in `v1beta1`, `v1beta2`, and `v1beta3` design, matching the real server struct's field names and json tags exactly. Left the `200` response (`ResourceAccessMapping`) and the sibling GET-by-actor-type operation untouched: I have no confirmed evidence for either shape, only for the request body issue #1144 describes.

Field naming choices, stated in the schema's own descriptions rather than assumed:

- `actorType` is a plain string, not an enum of one value. The server field itself is a plain Go `string` with no enforced enum; locking it to `user` would make the schema reject a legitimate future actor type before the server itself does.
- `ResourceAccessMappingPayload`'s three fields are all required, matching what the only confirmed-correct client actually sends on every request, not the Go struct's `omitempty` tags, which govern decode leniency, not what a well-formed request looks like.

Also documented, on the operation itself, why the generated RTK Query mutation cannot auto-invalidate the affected resource's list cache: the operation spans three resource types chosen at call time via a path parameter, and this repo's RTK codegen only supports static per-operation tags.

## Test plan

- [x] `make generate-golang`: generated struct field names/json tags match the server struct quoted in the issue exactly, across all three versions
- [x] `npm run generate:types`: generated TypeScript matches
- [x] `make generate-rtk`: `handleResourceShare` mutation generated correctly (the old, wrongly-pathed `shareResource` mutation is gone), RTK/gofmt regression suites pass
- [x] `go run ./cmd/validate-schemas`: no blocking violations

## Summary by CodeRabbit

- **Improvements**
  - Resource sharing requests now use a structured format for granting and revoking access, identifying actors by ID and type.
  - Notifications can be explicitly enabled or disabled as part of resource-sharing requests.
  - Resource-sharing models are now consistently typed across supported API versions.

- **Documentation**
  - Updated API documentation to clarify grant, revoke, notification, multi-resource behavior, and cache-refresh responsibilities.

- **Removed**
  - Removed the separate generated resource-sharing mutation and hook from the client API.


MU5A:
Checked this properly: the observation is accurate, `invalidatesTags: ["ResourceAccess_resource_access"]` is currently a no-op, no query anywhere provides that tag, so a successful share does not refresh any list right now.

The specific fix suggested (invalidate the matching design/filter/view tag based on `queryArg.resourceType`) isn't something this repo's generation pipeline can produce, though. RTK Query cache tags here are static per operation, derived from the operation's own OpenAPI `tags:` field (confirmed against `publishCatalogContent`, which gets its cross-resource `invalidatesTags: ["Design_designs"]` simply by being tagged `designs`, since it only ever affects one resource type). `shareResource` is one generic endpoint spanning three different resource types chosen at call time via the `resourceType` path parameter, and a static per-operation tag can't express "invalidate whichever tag matches this specific call." That would need a hand-written `invalidatesTags` callback, which isn't something any endpoint in this generated client has, and adding one just for this endpoint would be a much larger, precedent-setting change than this PR should carry.

Documented the limitation directly on the operation instead, so a consumer of `useShareResourceMutation` knows up front that they need to invalidate or refetch the affected list themselves after a successful share.

MU5A:
Correcting course on this one, flagging in case anyone already looked at the first version: the path I originally added (`/api/resources/{resourceType}/{resourceId}/share`, taken directly from issue #1144's own text) is wrong.

While checking CodeRabbit's cache-invalidation comment I found this repo already has an incomplete `handleResourceShare` operation at the real path, `/api/resource/{resourceType}/share/{resourceId}`, present identically in v1beta1/v1beta2/v1beta3 design, just with an untyped placeholder for its request/response. Confirmed against `meshery/meshery`'s current `ui/rtk-query/resource.ts`, which calls that exact path and whose own code comment describes this issue's bug verbatim.

Removed the wrong new construct and filled in the real, existing operation instead across all three versions. PR description rewritten to match, full details there.

PR #1216, opened 2026-09-14
Rename CatalogData.class/snapshotURL to match the real wire

### LINKED ISSUE #1142: CatalogData declares `class`/`snapshotURL`; every server emits `contentClass`/`imageURL`
## Summary

`CatalogData` (`schemas/constructs/v1beta2/catalog/catalog.yaml`, referenced by the v1beta3 design construct as `MesheryPattern.catalogData`) names two fields that no server ever emits, and omits the two that every server does emit. With `additionalProperties: false` on the same object, the canonical schema does not describe - and would reject - the payload that is actually on the wire.

## What the wire carries

Both servers marshal `catalogData` from meshkit's `models/catalog/v1alpha1.CatalogData`:

- `meshery-cloud`: `server/models/meshery_patterns.go` -> `CatalogData *v1alpha1.CatalogData \`json:"catalogData" db:"catalog_data"\``
- the same meshkit type backs the Meshery server side.

`meshkit/models/catalog/v1alpha1/catalog.go`:

```go
type CatalogData struct {
    ContentClass     ContentClass               `json:"contentClass,omitempty"`
    PublishedVersion string                     `json:"publishedVersion"`
    Compatibility    []CatalogDataCompatibility `json:"compatibility"`
    PatternCaveats   string                     `json:"patternCaveats"`
    PatternInfo      string                     `json:"patternInfo"`
    // Contains reference to the dark and light mode snapshots of the catalog.
    SnapshotURL []string `json:"imageURL,omitempty"` // ... updated the json tag to match previous key name, so changes will not be required in existing catalogs
    Type        CatalogDataType `json:"type"`
}
```

Note the explicit comment on `SnapshotURL`: the JSON tag was deliberately held at the legacy `imageURL` key so stored catalog rows would not need rewriting.

meshery-cloud queries the stored JSONB by that name too - `server/dao/meshery_patterns_dao.go`:

```go
// All rows backfilled to canonical `contentClass`
finalWhere, args := utils.GenerateWhereClause("LOWER(catalog_data->>'contentClass') = LOWER(?)", catalogClasses)
```

## What the canonical declares

`schemas/constructs/v1beta2/catalog/catalog.yaml` declares `class` and `snapshotURL`, and sets `additionalProperties: false`.

| wire key (meshkit, both servers) | canonical key |
| --- | --- |
| `contentClass` | `class` |
| `imageURL` | `snapshotURL` |

## Blast radius

1. Any consumer that types `catalogData` from `@meshery/schemas` reads `class` / `snapshotURL` and gets `undefined`. Sistent's catalog cards, catalog detail page and performers section all read `contentClass` / `imageURL` and are therefore correct against the wire and wrong against the canonical - see layer5io/sistent#1780, which keeps a deliberately narrower local `PatternCatalogData` pending this issue.
2. `additionalProperties: false` means a strict validator run against a real catalog payload rejects it on two unknown properties.

## Two related inconsistencies in the same construct

**a) The `class` enum disagrees with `contentClass` elsewhere in this repo.** `CatalogData.class` allows `official | verified | reference architecture`, while the catalog-content constructs generated into `constructs/v1beta3/design/Design.d.ts` and `cloudApi.d.ts` type `contentClass` as `official | verified | project | community`. Same concept, two enums, neither a subset of the other.

**b) The generated Go models disagree with the YAML.** `models/v1beta1/catalog/catalog.go` and `models/v1alpha2/catalog/catalog.go` tag `pattern_caveats` / `pattern_info` in snake_case, while the YAML (and the emitted TypeScript, and the wire) use `patternCaveats` / `patternInfo`. So this single construct currently has three disagreeing representations inside `meshery/schemas` itself.

## Suggested resolution

Rename `class` -> `contentClass` and `snapshotURL` -> `imageURL` in `constructs/v1beta2/catalog/catalog.yaml` to match the wire, reconcile the two `contentClass` enums, and regenerate so the Go models stop emitting `pattern_caveats` / `pattern_info`. If instead the intent is for the servers to move to `class` / `snapshotURL`, that needs a dual-accept window in meshkit first, because the stored `catalog_data` JSONB rows use the legacy keys and are queried by them.

MU5A:
Fixes #1142

`CatalogData` named two fields no server ever emits (`class`, `snapshotURL`) and omitted the two every server does emit (`contentClass`, `imageURL`), with `additionalProperties: false` meaning the canonical schema would reject the payload actually on the wire. Both meshery-cloud and Meshery server marshal `catalogData` from meshkit's `models/catalog/v1alpha1.CatalogData`, whose json tags are `contentClass` and `imageURL` (the `SnapshotURL` field is deliberately still tagged `imageURL` in Go, to avoid a rewrite of stored catalog rows).

## Scope: all three copies, not just the one named

Fixed in all three near-identical copies of this construct: `v1beta2/catalog` (the one the issue names, referenced by `MesheryPattern.catalogData`), `v1beta1/catalog`, and `v1alpha2/catalog`. This wasn't optional. `MesheryPattern.catalogData` carries an `x-go-type: catalogv1alpha2.CatalogData` override, so the actual Go struct backing it is generated from the **v1alpha2** copy, not the v1beta2 one. Fixing only the file the issue names would have left the real Go type still wrong.

## The enum, verified against the actual source

Also fixed the `class` enum itself, verified against meshkit's real `ContentClass` const block (`models/catalog/v1alpha1/catalog.go`): the real values are `official`, `verified`, `community`. The schema's previous enum (`official`, `verified`, `"reference architecture"`) matched neither that nor the differently-wrong four-value enum already in this repo for the unrelated `CatalogRequest.contentClass` (a separate Postgres column on `catalog_requests`, left untouched here since I have no way to verify its real allowed values). The corrected `community` entry uses the exact description text meshkit's own `GetCatalogClasses()` gives it, which was already sitting, unused, in the old field's own top-level description text, a strong tell that `community` was always the intended third value.

## Also fixed while in the file

`v1beta1/catalog`'s separate `pattern_caveats`/`pattern_info` snake_case drift (the YAML, and therefore the Go json tags, disagreed with v1alpha2 and v1beta2's `patternCaveats`/`patternInfo` for the same concept, the second inconsistency the issue named). Updated the affected templates and the publish form's description text so nothing in the construct still names the old fields.

## Test plan

- [x] `make generate-golang`: all three copies now agree; `v1beta3/design`'s `MesheryPattern.catalogData` (via the `x-go-type` override) confirmed to pick up the v1alpha2 fix
- [x] `npm run generate:types`: confirmed the renamed fields propagate through `pattern`, `workspace`, `design` (v1beta2 and v1beta3), and `filter`
- [x] `make generate-rtk`: RTK/gofmt regression suites pass
- [x] `go run ./cmd/validate-schemas`: no blocking violations
- [x] `go test ./validation/...`: passes, including the form-subset guard confirming `publish.json` still validates against the renamed canonical schema


Junnygram (approved):
LGTM

PR #1218, opened 2026-09-15
fix(organization_smtp): add unsupported auth mechanism outcome

### LINKED ISSUE #1196: organization_smtp: no outcome for "the server does not offer the selected auth mechanism"
### Summary

`OrganizationSmtpTestResultOutcome` (and its sibling `OrganizationSmtpConfigurationLastFailureReason`) cannot express "the mail server does not offer the authentication mechanism the organization selected". The nearest member, `auth_rejected`, states that the credential was refused — which sends an administrator to re-enter a password that is correct.

### How it is reachable

`authMechanism` is now honoured at dial time in meshery-cloud rather than assumed: selecting `cram-md5` dials CRAM-MD5 rather than silently downgrading to PLAIN, because the point of choosing it is that the password never crosses the wire.

`smtp.CRAMMD5Auth(...).Start` returns its mechanism name unconditionally — unlike `PlainAuth.Start` it inspects neither `ServerInfo.TLS` nor `ServerInfo.Auth` — so against a relay advertising only PLAIN/LOGIN the client sends `AUTH CRAM-MD5` and takes the server's 504. The server now reads `client.Extension("AUTH")` and refuses before sending, with prose that names the real problem and tells the operator to change the method. **The prose is right; only the typed outcome is imprecise.**

### Proposed addition

A member along the lines of `auth_mechanism_unsupported` on:

- `OrganizationSmtpTestResult.outcome`
- `OrganizationSmtpConfiguration.lastFailureReason`

Both enums already distinguish this class of "the server cannot do what was configured" from "the server refused what was presented" — `starttls_unsupported` is exactly that distinction for transport encryption, and is the precedent that this shape is accepted here. `auth_mechanism_unsupported` is its authentication counterpart.

### Not urgent, and deliberately not blocking

meshery-cloud reports `auth_rejected` for this case today, with a comment at the mapping site naming the gap and pointing here. The absence costs **precision, not function**: the operator receives the correct instruction in the human-readable message and can act on it. A consuming UI that branches on the outcome will show its `auth_rejected` copy alongside a message describing a mechanism problem, which is a mismatch worth closing but not a broken flow.

The downstream change once this lands is small: extend the mapping to the new member and update the UI's outcome-to-copy map.

piyushtripathi9424:
## Summary

- Add `auth_mechanism_unsupported` to Organization SMTP test outcomes.
- Add `auth_mechanism_unsupported` to the Organization SMTP last failure reason enum.
- Regenerate the downstream TypeScript contracts to include the new enum value.
- This distinguishes an unsupported authentication mechanism from rejected credentials, providing a more accurate failure classification.

## Validation

- `make validate-schemas` passes.
- Regenerated TypeScript contracts successfully.

## Summary by CodeRabbit

- **New Features**
  - Added `auth_mechanism_unsupported` as an SMTP test outcome and failure reason.
  - SMTP delivery failures can now indicate that the remote server does not support the configured authentication mechanism.
  - Regenerated the downstream TypeScript contracts to reflect the new enum value.

Fixes #1196


leecalcote (commented):
Thanks for working on this, @piyushtripathi9424

piyushtripathi9424:
> Thanks for working on this, @piyushtripathi9424

Thanks, Lee! Glad to work on this. I appreciate you taking the time to review it.

